PR #3, opened 2019-12-24
Better pinout file filter, ignore generated files, updated readme.md.

derekmulcahy:


readme.md
```diff
@@ -6,11 +6,13 @@ Project Apicula uses a combination of fuzzing and parsing of the vendor data fil
...
 
-Version 1.9 of the Gowin vendor tools, and a copy of the Excel pinout files downloaded in `~/Documents/gowinsemi`
+Version 1.9.1.01 of the Gowin vendor tools, and a copy of the Excel pinout files downloaded in `~/Documents/gowinsemi`:
+* UG107-1.07E_GW1N-1 Pinout.xlsx
+* UG801-1.5E_GW1NR-9 Pinout.xlsx
 
 The latest Yosys and Nextpnr, installed with the generic backend.
 
-Python 3.7+:
+Python 3.8:
```
pepijndevos: this is no longer true, should work on 3.6 now with the dataclasses module

readme.md
```diff
@@ -6,11 +6,13 @@ Project Apicula uses a combination of fuzzing and parsing of the vendor data fil
 
 ##  Dependencies
 
-Version 1.9 of the Gowin vendor tools, and a copy of the Excel pinout files downloaded in `~/Documents/gowinsemi`
+Version 1.9.1.01 of the Gowin vendor tools, and a copy of the Excel pinout files downloaded in `~/Documents/gowinsemi`:
+* UG107-1.07E_GW1N-1 Pinout.xlsx
+* UG801-1.5E_GW1NR-9 Pinout.xlsx
```
pepijndevos: I tested it with 1.9.3 now, so any 1.9 release should work.
Maybe small nitpick, but I'd write "copy of the following"
Also worth noting: I used an older version of the excel files, so the names change sometimes.

I'm a bit torn. Might be good to specify exact versions for reliability,  but it really should work with any recent version.

readme.md
derekmulcahy: I thought the start parameter in sum() was new in 3.8.

readme.md
derekmulcahy: I would recommend specifying what you use. If someone has luck with other versions then it's a secondary win.

readme.md
pepijndevos: Yes, so I'm no longer using it, see 2d8d2e7653bfbe33876f51f13603dd137990ae8d

PR #8, opened 2020-03-12
gowin_pack.py: convert bits vector to int before sum

Martoni:
My python can't sum string bit vector.

### DIFF
--- a/gowin_pack.py
+++ b/gowin_pack.py
@@ -50,7 +50,7 @@ def place(db, tilemap, bels):
                     tile[row][col] = 1
 
         elif typ == "GENERIC_IOB":
-            assert sum(attr.values()) <= 1, "Complex IOB unsuported"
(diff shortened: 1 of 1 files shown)


pepijndevos:
Huh, I don't remember the type of this attribute, but if it's not an int or boolean, we have to be careful that the lines below that test for their truth value actually work correctly.

Looking at #9 I have a suspicion that maybe your fix resulted in no IOB being generated at all. Could you verify `if attr["INPUT_USED"]:` does the correct thing? I'll have a closer look at this PR over the weekend.

Martoni:
Yes it was wrong type for attr["INPUT_USED"] !
Trabucayre fixed it in his [PR](https://github.com/pepijndevos/apicula/pull/10/commits/c39e56ce464ed0ebd8b7ead4317cbc0861079c5c)

pepijndevos:
Thank you! Looking forward to see any future cool projects and contributions!

Issue #9, opened 2020-03-12
LEDs doesn't blink


Martoni:
I followed the tutorial given in main README for the board TEC0117.
```
$ export GOWINHOME=/chemin/de/linstallation/gowin/IDE/
$ export DEVICE="GW1NR-9" # TEC0117
$ python dat19_h4x.py # makes $DEVICE.json
$ python tiled_fuzzer.py # makes $DEVICE.pickle
$ cd generic
$ bash simple.sh blinky.v # TEC0117
$ cd ..
$ python gowin_pack.py generic/pnrblinky.json
$ python gowin_unpack.py pack.fs
$ yosys -p "read_verilog -lib +/gowin/cells_sim.v; clean -purge; show" unpack.v
$ openFPGALoader -m -b littleBee pack.fs # FOSS programmer
```

 The bitstream pack.fs has been generated and downloaded with openFPGALoader but LEDs doesn't blink. All LED are on.

I used official «pinout» from Gowin website maybe it's that ?

I have an error after downloading the bitstream but I read on [this issue](https://github.com/trabucayre/openFPGALoader/issues/14) that is a known bug.

pepijndevos:
Can you share the output of the unpack and show steps? Show can be told to output a PNG or SVG rather than show a window. The key thing to look for is if the inputs are fully connected to the outputs.

If there is nothing to render, this is a very bad sign. Remove the purge argument and try again. This will produce a HUGE graph. You can look at the source code and search for the IOB locations. As mentioned, I have a hunch that your IOB did not get created correctly.

You can also pass a selector to show, to only show the IOB, or you can of course just look at unpack.v and search around.

Martoni:
Solved by trabucayre [PR](https://github.com/pepijndevos/apicula/pull/10) !

pepijndevos:
Cool, so you have a working blinky! Yay!

Martoni:
https://www.youtube.com/watch?v=yc2yBYlCCI4&feature=youtu.be

\o/

PR #10, opened 2020-03-12
fix generic_iob direction detection

trabucayre:
for GENERIC_IOB type, attr is a dict where keys are "INPUT_USED" and "OUTPUT_USED" and values are number of input/output in string (binary representation).
Since values are string, tests:
```
if attr["INPUT_USED"]:
[...]
elif attr["OUTPUT_USED"]:
```
are based on empty/None instead of 0 or > 0.
So, first test is always true and output iob are generate as input too.

fix issue [#9](https://github.com/pepijndevos/apicula/issues/9)
PR [#8](https://github.com/pepijndevos/apicula/pull/8) must be applied too.

### DIFF
--- a/gowin_pack.py
+++ b/gowin_pack.py
@@ -52,9 +52,9 @@ def place(db, tilemap, bels):
         elif typ == "GENERIC_IOB":
             assert sum(attr.values()) <= 1, "Complex IOB unsuported"
             iob = tiledata.bels[f'IOB{num}']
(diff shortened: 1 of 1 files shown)


pepijndevos:
Thank you! Quick debugging work :) I checked with David and Nextpnr indeed changed the data types on those. Looking forward to see any cool projects or contributions you might be working on :)

PR #11, opened 2020-03-22
gowin_pack: introduce the configuration data checksum computation

trabucayre:
Current configuration data checksum (l.1 in header, l2 in footer) is hardcoded.
The checksum, as mentionned in TN653 2.2.7, is an accumulation of consecutive 16bits.
This patch introduce this algorithm to update db.cmd_hdr and db.cmd_ftr.

### DIFF
--- a/gowin_pack.py
+++ b/gowin_pack.py
@@ -2,6 +2,7 @@ import sys
 import os
 import re
 import pickle
+import numpy as np
 import json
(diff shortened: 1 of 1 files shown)


pepijndevos:
I was completely unaware this was just documented in the open. Great work! Thanks a lot :)

PR #13, opened 2020-08-02
Avoid adding duplicate wires

kbeckmann:
The current solution where exceptions are caught doesn't seem to work with newer versions of nextpnr. This issue is avoided by keeping track of added wires and not adding them at all instead.

### DIFF
--- a/generic/simple.py
+++ b/generic/simple.py
@@ -16,14 +16,17 @@ if not device:
 with open(f"../{device}.pickle", 'rb') as f:
     db = pickle.load(f)
 
+
+added_wires = []
 def addWire(row, col, wire):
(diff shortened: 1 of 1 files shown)


pepijndevos:
Thank you!

generic/simple.py
```diff
@@ -16,14 +16,17 @@
 with open(f"../{device}.pickle", 'rb') as f:
     db = pickle.load(f)
 
+
+added_wires = []
```
smunaut, line 20: Just as a random unsolicited advice, using a set() would probably be faster ...

kbeckmann: Thanks, indeed it boost performance a lot. Didn't know this was such hot code :).

PR #14, opened 2020-08-03
Add github workflow to build and upload chipdb

kbeckmann:
This will build the chipdb for each new commit to the master branch and upload it as a github artifact . Not sure if that is a bit redundant, but at least it runs automatically whenever stuff changes.

An example run of this can be seen in https://github.com/kbeckmann/apicula/actions/runs/193110067

The reason to why I changed `ADD` to `RUN curl..` is so that it gets cached. Docker will download the files every single time the container is built otherwise.

### DIFF
--- a/.github/workflows/chipdb.yml
+++ b/.github/workflows/chipdb.yml
@@ -1,0 +1,22 @@
+name: Chipdb builder
+
+on:
+  push:
+    branches:
+        - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build chipdb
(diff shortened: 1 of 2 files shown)


kbeckmann:
I updated the URL for the GW1NR-9 pinout file, since the old URL returned 0 bytes. I guess this means that it has been updated, which is the reason to why the tools don't work with it anymore with the new file.

kbeckmann:
Updated the PR with new URLs. Builds GW1NR-9 fine now. Example artifacts: https://github.com/kbeckmann/apicula/actions/runs/193294274

pepijndevos:
Amazing!

PR #17, opened 2020-09-17
librarify the top of the dat parser

NotAFile:


NotAFile:
Okay I instantly got distracted with something else, maybe next time.

PR #20, opened 2020-10-03
generic/simple.py: force wire type to be a string

trabucayre:
Some wire name are composed only to digit ( for instance `1007`).
python, in this case, uses a `int` type instead of `string`. The consequences are failure in some call
for instance in:
```
def wire2global(row, col, db, wire):
    if wire.startswith("G") or wire in {'VCC', 'VSS'}:
```
where `int` as no `startwith`

### DIFF
--- a/generic/simple.py
+++ b/generic/simple.py
@@ -33,7 +33,7 @@ for row, rowdata in enumerate(db.grid, 1):
         # add wires
         wires = set(chain(tile.pips.keys(), *tile.pips.values()))
         for wire in wires:
-            addWire(row, col, wire)
+            addWire(row, col, str(wire))
         # add aliasses
         # creat bels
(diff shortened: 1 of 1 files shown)


pepijndevos:
When does this come up? 1000+ wires have an unknown function and are not used at all as far as I know. Before I "folded" them to a lower number because the 1000 seems to be more of a flag. I guess I should just revert that change rather than this patch, which doesn't address the reason these wires show up.

pepijndevos:
Reverted the change in 583b67a00e7c35c4a6dc402099c656245e1adc10

trabucayre:
Ok.
Finally, this PR is more or less "defensive" to force python to use the correct type. But if this aspect is asserted previously It's good too (and maybe more easy).

PR #22, opened 2020-12-29
partial complement of the fs header

trabucayre:
This PR try to complete header line content description and to clarify bit/area unknown for futher analyze.


pepijndevos:
Awesome!

PR #25, opened 2021-01-09
Fix CRC check and errors when using python 3.8.5

roman3017:
I had to make the following changes to remove problems on my system Ubuntu 20.04.1 LTS:

 - set axis argument to None/default in np.packbits
 - remove usage of importlib.resources
 - make gowin_* files executable
 - add parent to path to import from apycula

With these changes I can compile nextpnr-gowin, make this project, and also openFPGALoad does not complain about CRC failures anymore.

```sh
# tested with:
cd apicula/examples/
make
nextpnr-gowin --json blinky.json --write pnrblinky.json --device GW1NR-LV9QN881C6/I5 --cst tec0117.cst
../apycula/gowin_pack.py -d GW1N-9 -o blinky.fs pnrblinky.json
openFPGALoader -b littleBee -m blinky.fs
#openFPGALoader -b littleBee -r -f blinky.fs
```

### DIFF
--- a/apycula/gowin_pack.py
+++ b/apycula/gowin_pack.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 import sys
 import os
 import re
@@ -5,7 +6,7 @@ import pickle
 import numpy as np
 import json
 import argparse
-import importlib.resources
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), '..'))
 from apycula import chipdb
 from apycula import bslib
 from apycula.wirenames import wirenames, wirenumbers
@@ -27,6 +28,8 @@ def get_pips(data):
             if res:
                 row, col, src, dest = res.groups()
                 yield int(row), int(col), src, dest
+            elif "ALIAS" in pip:
+                print("Ignoring alias:", pip)
             elif pip:
                 print("Invalid pip:", pip)
 
@@ -108,7 +111,7 @@ def header_footer(db, bs):
     CRC_check and security_bit_enable set
     """
     bs = np.fliplr(bs)
-    bs=np.packbits(bs, axis=1)
+    bs = np.packbits(bs)
     # configuration data checksum is computed on all
     # data in 16bit format
     bb = np.array(bs.flat)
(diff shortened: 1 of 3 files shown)


pepijndevos:
Thanks for the PR!

Fixing the CRC is great. However, your other changes seem misguided, based on trying to run the scripts directly rather than letting pip create binaries. When using the setup.py generated scripts, the paths are set up correctly, which is not the case when you try to run the scripts directly. You can actually run them with `python -m apycula.gowin_pack`.

Please do a PR with only the CRC fixes, without altering the intended installation functionality.

If the current readme instructions are unclear, I'd happily also accept a separate PR that clarifies that the python binary path should be added to `$PATH` which is not the default on some distributions, which could be the source of your problem running the code as intended.

roman3017:
Ok, will close this PR and will think about how to get it working without adding paths to script, which currently does not work for me. Likely my setup is different from expected.

pepijndevos:
Feel free to hop into #apicula on Freenode IRC if you want to debug your issue and discuss improvements. I want to make installation as easy as possible so evidently there is room for improvement.

roman3017:
One more thing, with this change one can build nextpnr with:
```sh
cmake . -DARCH="gowin;generic" -DGOWIN_BBA_EXECUTABLE=$PWD/../apicula/apycula/gowin_bba.py
make
sudo make install
```
The command `python -m apycula.gowin_pack` works when packing pnrblinky.json. I wonder if there is a similar trick here.

pepijndevos:
So, if you're using a virtualenv, then the path would be `path/to/env/bin/gowin_bba`. On my system, pip install binaries into `~/.local/bin`, so if that is the case, but is not on your path, you either need to run `export PATH="~/.local/bin:$PATH"` or you need to build nextpnr with `cmake . -DARCH="gowin" -DGOWIN_BBA_EXECUTABLE=~/.local/bin/gowin_bba`

apycula/gowin_pack.py
```diff
@@ -27,6 +28,8 @@ def get_pips(data):
             if res:
                 row, col, src, dest = res.groups()
                 yield int(row), int(col), src, dest
+            elif "ALIAS" in pip:
+                print("Ignoring alias:", pip)
```
pepijndevos, line 32: Huh, did you actually run into aliases with the nextpnr-gowin flow? AFAIK this was only a nextpnr-generic thing.

roman3017: You are right, I added this when using generic and it was bit confusing to see errors.

pepijndevos: I doubt that would have worked anyway because the primitives have different names. It wouldn't be that hard *at the moment* to support both, until support for more primitives is added.

PR #26, opened 2021-01-10
fix crc check fail when using python 3.8.5

roman3017:
- set axis argument to None/default in np.packbits
 
With this patch I do not see `CRC check : FAIL` anymore when flashing with openFPGALoad. Tested with:

```sh
cd examples/
rm -rf *.json
make
nextpnr-gowin --json blinky.json --write pnrblinky.json --device GW1NR-LV9QN881C6/I5 --cst tec0117.cst
gowin_pack -d GW1N-9 -o blinky.fs pnrblinky.json
openFPGALoader -b littleBee -r -f blinky.fs 
```

### DIFF
--- a/apycula/gowin_pack.py
+++ b/apycula/gowin_pack.py
@@ -108,10 +108,10 @@ def header_footer(db, bs):
     CRC_check and security_bit_enable set
     """
     bs = np.fliplr(bs)
-    bs=np.packbits(bs, axis=1)
+    bs=np.packbits(bs)
     # configuration data checksum is computed on all
     # data in 16bit format
-    bb = np.array(bs.flat)
+    bb = np.array(bs)
 
     res = int(bb[0::2].sum() * pow(2,8) + bb[1::2].sum())
     checksum = res & 0xffff


apycula/gowin_pack.py
```diff
@@ -108,7 +108,7 @@ def header_footer(db, bs):
     CRC_check and security_bit_enable set
     """
     bs = np.fliplr(bs)
-    bs=np.packbits(bs, axis=1)
+    bs=np.packbits(bs)
     # configuration data checksum is computed on all
     # data in 16bit format
     bb = np.array(bs.flat)
```
pepijndevos, line 114: Am I correct in thinking that this `flat` is unneeded now, because `packbits` without axis arguments already flattens the array?

pepijndevos:
The reason this fix works is actually interesting. It makes no difference on GW1N-1, but fixes the problem on GW1N-9. I think this is because the width of a row is not a multiple of 16, and apparently they are just wrapped around, while previously they were calculated per row, adding some padding to each row basically.

pepijndevos:
Thank you very much for the PR! I take it you got it to work in the end? Do you think a clarification in the readme would be in order, so others don't fall in the same trap as you?

roman3017:
Yes, I got it working using `pip install -e .` and `-DGOWIN_BBA_EXECUTABLE="~/.local/bin/gowin_bba"` for nextpnr. Thank you very much for your help and clarification.

pepijndevos:
Thanks. I added a note to the readme to clarify adding the binaries to the path or passing their paths to nextpnr.

PR #27, opened 2021-01-10
Update build steps in Makefile

roman3017:
- add steps to build all fs files by default
 - add clean
 - add prog to program blinky

This change enables to run blinky test on TEC0117 with:
```sh
make clean
make
make prog
```

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,7 +1,26 @@
-all: attosoc.json blinky.json
+DEVICE ?= GW1N-9
+PN ?= GW1NR-UV9QN881C6/I5
+BOARD ?= tec0117
+
+all: attosoc.fs blinky.fs
+
+%.fs: %.pnr.json
+	gowin_pack -d ${DEVICE} -o $@ $^
+	cp $@ ${BOARD}-$@
+
+%.pnr.json: %.json
+	nextpnr-gowin --json $^ --write $@ --device ${PN} --cst ${BOARD}.cst
 
 attosoc.json: attosoc/attosoc.v attosoc/picorv32.v
 	yosys -p "synth_gowin -json $@" $^
 
 %.json: %.v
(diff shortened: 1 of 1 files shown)


examples/Makefile
```diff
@@ -1,7 +1,22 @@
...
+DEVICE ?= GW1N-9
+PN ?= GW1NR-LV9QN881C6/I5
+
+all: attosoc.fs blinky.fs
+
+%.fs: %.pnr.json
+	gowin_pack -d ${DEVICE} -o $@ $^
+
+%.pnr.json: %.json
+	nextpnr-gowin --json $^ --write $@ --device ${PN} --cst tec0117.cst
```
pepijndevos, line 10: The `tec0117.cst` still hardcodes it to that particular board

pepijndevos:
I like the idea of this change, and it's something I was planning to anyway.

I was thinking if you encode the device in the bitstream name, so you can just `make blinky-GW1N-4.fs` and it'd do the right thing. I think you could just have a rule thats `%-GW1N-4.fs` that passes the right part number and cst file?

examples/Makefile
```diff
@@ -1,7 +1,22 @@
...
 	yosys -p "synth_gowin -json $@" $^
 
 %.json: %.v
 	yosys -p "synth_gowin -json $@" $^
+
+prog:
+	openFPGALoader -b littleBee -r -f blinky.fs
+
+clean:
+	rm -f *.json *.fs
```
pepijndevos, line 22: I think it's good style to make these targets phony?

examples/Makefile
roman3017: Ok, I will make the change.

examples/Makefile
roman3017: Ok, I will make the change.

PR #28, opened 2021-01-28
readme, examples/Makefile: fix openFPGALoader target board name

trabucayre:
Flash instruction and Makefile rules update:
- instead of using the ambiguous *littleBee* name a new entry *tec0117* has been added
- seeestudio runber is now officially supported


pepijndevos:
Amazing, thanks!

PR #29, opened 2021-02-13
compress algorithm support

trabucayre:
bitstream compression is more or less the replacement of an serie `0x00` by a key.

this PR adds:
- CLI arg to enable compression
- header update with keys
- function to compress line
- a draft of algorithm's documentation

Tested with all devices supported by apicula


pepijndevos:
Holy shit programming a blinky is fast now. Thanks so much!

PR #30, opened 2021-04-27
Make it clearer what the example values are from

cincodenada:
I came trying to build for a RUNBER and was confused about what tec0117 was but put it in anyway, which obviously didn't work, ha! Figured it's be useful to be more explicit here.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -39,7 +39,7 @@ nextpnr-gowin --json blinky.json \
               --write pnrblinky.json \
               --device GW1NR-UV9QN881C6/I5 \
               --cst tec0117.cst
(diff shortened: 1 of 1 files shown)


pepijndevos:
Why did you change "seems to us" though? The Runber DEFINITELY uses GW1N-4, no doubt about it.

cincodenada:
Oh great, I'll change it back then - I just wasn't sure myself :)

pepijndevos:
Actually, maybe it would be even clearer to use full part numbers everywhere? This is actually supported
https://github.com/YosysHQ/apicula/blob/02421adb720ba1e4ffce5ec625279d5489e7b3ad/apycula/gowin_pack.py#L141-L142

So I think it would be best to remove the comment, and use `GW1NR-UV9QN881C6/I5` instead of `GW1N-9` and explain in the paragraph that the user needs to change tec0117 and the part number.

PR #33, opened 2021-07-05
examples: fix the number of leds according to the targeted board

trabucayre:
By default the number of leds is hardcoded to 8. But depending to the targeted board this value is not always true and randoms pins are used for physical output not availables.

This PR add a verilog define to pass real number of leds:
- now it's no more possible to use a generic blinky.json to address all boards -> rule is renamed to append expliciitly board name and `synth` keyword
- the number of leds is provided trough Makefile variables

To clarify way to obtain number of leds:
```make
yosys -D LEDS_NR=$($(shell echo $(@) | cut -d\- -f 2)_leds) -p "synth_gowin -json $@" $^
```
is a two steps replacement:
1. `echo $(@) | cut -d\- -f 2` split rule name using '-' as separator and extract 2nd field (board name)
2. now `LEDS_NR=$($(shell echo $(@) | cut -d\- -f 2)_leds)` become `LEDS_NR=$(tangnano_leds)` (for tangano)
3. this new variable is replaced by the value 3

This patch drop reg to the entity section since led is combinatorial

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,31 +1,36 @@
+runber_leds=8
+tangnano_leds=3
+tec0117_leds=8
+
 all: attosoc-tec0117.fs blinky-tec0117.fs blinky-runber.fs blinky-tangnano.fs
 
 %-tec0117.fs: %-tec0117.json
 	gowin_pack -d GW1N-9 -o $@ $^
 
-%-tec0117.json: %.json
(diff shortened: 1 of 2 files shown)


pepijndevos:
The makefile hack is a bit evil but alright. Great to not have random pins blinking.

PR #34, opened 2021-07-05
Support gw1ns 2c

trabucayre:
This PR add support for GW1NS-2C FPGA:
The first commit avoid a crash like:
```bash
  File "/xxx/apicula/apycula/gowin_pack.py", line 91, in place
    bits = tiledata.bels['BANK'].modes['DEFAULT']
KeyError: 'BANK'
make: *** [Makefile:27 : blinky-honeycomb.fs] Erreur 1
```
Second commit add everything required to generate the pickle for this device and update readme (Dockerfile is partially updated -> needs to store xls somewhere)
And finally last commit add the honeycomb in examples

### DIFF
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -143,6 +143,9 @@ def get_pins(device):
     elif device == "GW1NR-9":
         header = 1
         start = 7
+    elif device == "GW1NS-2C":
+        header = 0
+        start = 7
     else:
         raise Exception("unsupported device")
     pkgs = pindef.all_packages(device, start, header)
@@ -164,6 +167,10 @@ def xls_pinout(family):
     elif family == "GW1N-4":
         return {
             "GW1N-4": get_pins("GW1N-4"),
+        }
+    elif family == "GW1NS-2":
+        return {
+            "GW1NS-2C": get_pins("GW1NS-2C"),
         }
     else:
         raise Exception("unsupported device")
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -143,6 +143,9 @@ def get_pins(device):
     elif device == "GW1NR-9":
         header = 1
         start = 7
+    elif device == "GW1NS-2":
+        header = 0
+        start = 7
     elif device == "GW1NS-2C":
         header = 0
         start = 7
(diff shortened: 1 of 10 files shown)


apycula/chipdb.py
```diff
@@ -165,6 +168,10 @@ def xls_pinout(family):
         return {
             "GW1N-4": get_pins("GW1N-4"),
         }
+    elif family == "GW1NS-2":
+        return {
+            "GW1NS-2C": get_pins("GW1NS-2C"),
```
pepijndevos, line 173: No `C` here, right?

apycula/tiled_fuzzer.py
```diff
@@ -31,6 +31,12 @@
 device = sys.argv[1]
 
 params = {
+    "GW1NS-2": {
+        "package": "QN48",
```
pepijndevos, line 35: Usually it's best to take the package with the most pins here, this speeds up fuzzing.

pepijndevos:
I have uploaded https://wishfulcoding.nl/gowin/UG825-1.2.1E_GW1NS-2C%20Pinout.xlsx and https://wishfulcoding.nl/gowin/UG822-1.2.1E_GW1NS-2%20Pinout.xlsx so please add these to the dockerfile.

Awesome work!!

apycula/chipdb.py
pepijndevos: Or rather... I think we should add both the C and non-C variants. They can be supported by the same chipdb, it's only the pins that get added.

pepijndevos: Like with the GW1N-9 and GW1NR-9, same bitstream, different pinouts.

trabucayre: There is no GW1NS-2C directory in IDE/share so yes need to use GW1NS-2
Pushed GW1NS-2 variant for pinout

apycula/tiled_fuzzer.py
trabucayre: Done

trabucayre:
Dockerfile updated too
Thanks

pepijndevos:
Great!! I'll do a release once CI is done.

PR #35, opened 2021-07-07
Pin modes

yrabbit:
Option of detecting and using the bits of the operating mode flags.


apycula/gowin_pack.py
```diff
@@ -15,7 +15,7 @@ def get_bels(data):
     for cell in data['modules']['top']['cells'].values():
         bel = cell['attributes']['NEXTPNR_BEL']
         row, col, num = belre.match(bel).groups() 
-        yield (cell['type'], int(row), int(col), num, cell['parameters'])
+        yield (cell['type'], int(row), int(col), num, cell['parameters'], cell['attributes'])
```
pepijndevos, line 18: This is getting a bit out of hand. Maybe we should make a named tuple or dataclass? Named tuple is probably the least refactoring. https://docs.python.org/3/library/collections.html#collections.namedtuple

pepijndevos:
I think the error log parsing is a great addition, but I'd like to use it differently.
I would refactor the fuzzer to not generate invalid combinations, and throw an exception if anything fishy shows up in the error log.

This way you don't waste resources fuzzing invalid combinations, and you make the fuzzer more robust by throwing an error on unexpected warnings.
This means hardcoding that an input doesn't have a slew rate, but that just seems logical.

pepijndevos:
We don't have much automated testing, mostly because it requires hardware in the loop, but I think having an example that demonstrates these features would be helpful.

apycula/tiled_fuzzer.py
```diff
@@ -253,16 +293,19 @@ def run_pnr(mod, constr, config):
...
         subprocess.run([gowinhome + "/IDE/bin/gw_sh", tmpdir+"/run.tcl"])
-        # print(tmpdir); input()
+        #print(tmpdir); input()
         try:
+            attrs = constr.attrs;
             return (*bslib.read_bitstream(tmpdir+"/impl/pnr/top.fs"), \
                    list(read_posp(tmpdir+"/impl/pnr/top.posp")), \
-                   config)
+                   config, attrs, \
+                   read_err_log(tmpdir+"/impl/pnr/top.log"))
```
pepijndevos, line 304: Actually I was thinking of this one. Or maybe both could be named tuples. Or not, it's not super important.

pepijndevos:
It seems like the IOB mode does not affect the flags, like `'IBUF&PULL_MODE=NONE': {(3, 24)}, 'OBUF&PULL_MODE=NONE': {(3, 24)}, 'IOBUF&PULL_MODE=NONE': {(3, 24)}` so I would suggest omitting the mode from the flags.  This would probably also simplify the code and reduce runs.

You can basically fuzz the input flags on IBUF and the output flags on OBUF and then just support both on IOBUF.

yrabbit:
Yes, it will reduce the time of base generation significantly. Although I must note that so far I do not see how this will reduce the code: the bit string for IOBUF mode will have to get at least once, which means a special case in the loop.

And the second point: now gowin_pack automatically ignores invalid flags for the mode because such information is in the database, if the mode is removed from the flags, then the correctness will have to check programmatically.

But in principle it is not difficult to do.

pepijndevos:
Fair. Do what you think is best.

apycula/gowin_pack.py
yrabbit: done

apycula/tiled_fuzzer.py
yrabbit: done

pepijndevos:
Looks awesome! Doing some testing and then it's good to merge I think.

Actually I should make CI run on PRs.

PR #41, opened 2021-07-29
Put log parsers in module

yrabbit:
The IOB fuzzer is likely to be a separate module so the general functions should be taken out beforehand.


yrabbit:
Forget about it. Doesn't fit the accepted approach:)

PR #42, opened 2021-08-02
IOBUF + TBUF fuzzer

yrabbit:
fuzz IOBUF by connecting OEN and OUT as pepijndevos adviced. TBUF by add DFF.


pepijndevos:
So is TBUF actually different from IOBUF? I was hoping that with the O->OE trick it would not need a separate fuzzer as it doesn't conflict with the DFF fuzzer.

yrabbit:
It is exactly the same as IOBUF, but this fact cannot be detected by your trick.
You want to take TBUF = IOBUF with a hanging O ?
This can be done by throwing out its fuzzling, implement TBUF in gowin_pack.

yrabbit:
are you sure that merging the fuzzers back into one file will be good? We won't have to separate them again when and if we get to merge registers and IOBs into one cell?

pepijndevos:
Generally my idea is that anything that can reasonably be fuzzed within tile boundaries should be in tiled_fuzzer, and things that need to be fuzzed across cell boundaries like clock wires need to be their own step.

So if it's the case that to fuzz IOB we need to put support structures in the tiles around it, then yes it can be a separate fuzzer. If the self-connected hack allows us to fuzz the IOB in a self-contained way, then it will be more efficient to merge it with other runs. This will basically mean that fuzzing PLLs and BRAM will incur no additional runtime since they can be in the same PnR runs as DFF and IOB.

I'm not sure I understand your comment about having to split up for registers in IOB. Again, if it is possible to fuzz registers in IOB in a self-contained way, it can be done within the tiled_fuzzer framework. But if you need to hook up the register to a LUT in an adjacent tile for it to work, then you are correct that it's better to do IOB fuzzing in a separate step.

In that case maybe we should refactor a bit so that all the common fuzzer stuff is in tiled_fuzzer and the specific ones can import that. Rather than the DFF fuzzer being in tiled_fuzzer and the other ones importing it.

pepijndevos:
I have not looked at the register-in-iob support at all. If it's just one bit that's basically an additional flag on IOB then I guess it can be fuzzed as such. If it's a full DFF that's independent of the IOB then it might be a seperate pass, which could still happen inside tiled_fuzzer. Depends if the tools will just delete the register if it's unconnected if we can fuzz it in a self-contained way I guess.

yrabbit:
Got it. 
I merge the fuzzers back together and throw out the extra pieces with TBUF and routing. I'll have to fix gowin_{pack, unpack} as well.
About the registers: so far I assume they are special and one of the ins/outs is necessarily connected to the IOB.

PR #45, opened 2021-08-29
Bram support

pepijndevos:


apycula/tiled_fuzzer.py
```diff
@@ -28,8 +28,8 @@
 if not gowinhome:
     raise Exception("GOWINHOME not set")
 
-# device = os.getenv("DEVICE")
-device = sys.argv[1]
+device = os.getenv("DEVICE")
+# device = sys.argv[1]
```
pepijndevos, line 32: Reminder that I'm an idiot.
I went through the trouble of `git add -p` and then just muscle memoried `git commit -am`

uis246:
Any news?

PR #57, opened 2021-11-04
Get info about packages/pins from the IDE files.

yrabbit:

### DIFF
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -249,7 +249,7 @@ def get_pins(device):
     return (pkgs, res, res_bank_pins)
 
 # returns ({partnumber: (package, device)}, {pins}, {bank_pins})
-def xls_pinout(device):
+def json_pinout(device):
     if device == "GW1N-1":
         pkgs, pins, bank_pins = get_pins("GW1N-1")
         return (pkgs, {
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -70,6 +70,7 @@ class Device:
     # a grid of tiles
     grid: List[List[Tile]] = field(default_factory=list)
     timing: Dict[str, Dict[str, List[float]]] = field(default_factory=dict)
+    packages: Dict[str, Tuple[str, str]] = field(default_factory=dict)
     pinout: Dict[str, Dict[str, Dict[str, str]]] = field(default_factory=dict)
     pin_bank: Dict[str, int] = field(default_factory = dict)
     cmd_hdr: List[ByteString] = field(default_factory=list)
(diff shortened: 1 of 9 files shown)


yrabbit:
Well, we'll see how the CI goes. Should break on examples because nextpnr is old, but it will create bases.

apycula/tiled_fuzzer.py
```diff
@@ -33,25 +33,22 @@
...
-        "package": "PG256",
-        "header": 0,
+        "package": "PBGA256",
         "device": "GW1N-4-PBGA256-6",
         "partnumber": "GW1N-LV4PG256C6/I5",
     },
     "GW1N-1": {
-        "package": "LQ144",
+        "package": "LQFP144",
         "header": 1, # stupid note in excel
```
pepijndevos, line 52: this header can go

apycula/chipdb.py
```diff
@@ -234,51 +235,57 @@ def from_fse(fse):
...
+    for pkg_rec in pkgs.values():
+        pkg = pkg_rec[0]
+        if pkg in res:
+            continue
+        res[pkg] = pindef.get_pin_locs(device, pkg, pindef.VeryTrue)
+        res_bank_pins.update(pindef.get_bank_pins(device, pkg))
+    return (pkgs, res, res_bank_pins)
+
+# returns ({partnumber: (package, device)}, {pins}, {bank_pins})
+def xls_pinout(device):
```
pepijndevos, line 252: maybe this shouldn't be called xls when it's not excel based ;)

pepijndevos:
Exactly as predicted! So what's up with the nextpnr side? I mean, isn't the chipdb format the same?

apycula/gowin_bba.py
```diff
@@ -147,6 +147,17 @@ def write_timing(b, timing):
     b.u32(len(timing))
     b.ref(blk)
 
+def write_partnumber_packages(b, db):
+    import ipdb; ipdb.set_trace()
```
pepijndevos, line 151: uh breakpoints should go I think haha

yrabbit: > Exactly as predicted! So what's up with the nextpnr side? I mean, isn't the chipdb format the same?

The structure has changed: a table with packages has been added.
You can put the next patch on nexpnr

https://github.com/yrabbit/nextpnr/commit/0e8a2999bda012175c1734abbf4efb76b515746c

yrabbit:
The speed is determined from the partnumber, the match is stored in the database so the input regular expression is simplified and allows more models:)

Of course this is only friendly with the last commit for nextpnr:  https://github.com/yrabbit/nextpnr/commit/74b4f6972878d90c6d5c22450dae535559edbea2

pepijndevos:
Looks good to me!

pepijndevos:
Wait actually if this is an incompatible bba change it should bump the DB version

pepijndevos:
https://github.com/YosysHQ/apicula/blob/b1dedc05ed4bf7ce622bd2e905c74867ac5331e6/apycula/gowin_bba.py#L186

examples/Makefile
```diff
@@ -49,7 +49,7 @@ nanolcd-tangnano-synth.json: nanolcd/TOP.v nanolcd/VGAMod.v
 	$(YOSYS) -p "synth_gowin -json $@" $^
 
 %-tec0117-synth.json: %.v
-	$(YOSYS) -D LEDS_NR=8 -p "synth_gowin -json $@" $^
+	$(YOSYS) -D LEDS_NR=4 -p "synth_gowin -json $@" $^
```
pepijndevos, line 52: Why did the number of LEDs change? :|

PR #62, opened 2021-11-09
Use LEDS_NR in the shift.v example

yrabbit:


examples/blinky.v
```diff
@@ -12,6 +12,6 @@ always @(posedge clk)
 
 // Combinational code (boolean logic)
 assign ctr_d = ctr_q + 1'b1;
-assign led = ctr_q[25:25-`LEDS_NR];
+assign led = ctr_q[25:25-(`LEDS_NR - 1)];
```
pepijndevos, line 15: this melts my brain haha

yrabbit: The good old off-by-one. Fun things start to discover when gowin_pack starts swearing about pinless IO:)

PR #67, opened 2021-12-08
Fix base modification.

yrabbit:
Python does not copy structures on assignment :(
The error with bank attributes was harmless, with IO attributes not so much.


pepijndevos:
Ouch

PR #68, opened 2021-12-10
Document the longval tables

yrabbit:
This is a very short description of what is found. IOB only.


pepijndevos:
This seems to reference your local changes to tiled_fuzzer.py but that'll be fixed eventually I suppose.

yrabbit:
That's right, I'll do a PR of the code when the testing is done.

PR #77, opened 2022-01-23
initial support for GW1N-9C

Icenowy:
Its bitstream structure seems to be the same with GW1N-9, according to
openFPGALoader code.

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ endif
 .SECONDARY:
 .PHONY: all clean
 all: apycula/GW1N-1.pickle apycula/GW1N-9.pickle apycula/GW1N-4.pickle \
-	 apycula/GW1NS-2.pickle apycula/GW1NS-4.pickle
+	 apycula/GW1NS-2.pickle apycula/GW1NS-4.pickle apycula/GW1N-9C.pickle
 
 %.json: apycula/dat19_h4x.py
 	python3 -m apycula.dat19_h4x $*
--- a/apycula/tiled_fuzzer.py
+++ b/apycula/tiled_fuzzer.py
@@ -89,6 +89,12 @@ params = {
         "partnumber": "GW1N-LV9PG256C6/I5",
         "recode_idx": recode_idx_gw1n9,
     },
+    "GW1N-9C": {
+        "package": "UBGA332",
+        "device": "GW1N-9C-UBGA332-6",
+        "partnumber": "GW1N-LV9UG332C6/I5",
+        "recode_idx": recode_idx_gw1n9, # TODO: recheck
+    },
     "GW1N-4": {
         "package": "PBGA256",
         "device": "GW1N-4-PBGA256-6",
--- a/.github/workflows/chipdb.yml
+++ b/.github/workflows/chipdb.yml
@@ -47,6 +47,24 @@ jobs:
       with:
           name: gw1n-9-stage
           path: GW1N-9*
+  gw1n9c:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build gw1n-9c chipdb
+      run: |
+        docker pull pepijndevos/apicula:1.9.8
+        docker run -v $(pwd):/usr/src/apicula pepijndevos/apicula:1.9.8 make apycula/GW1N-9C.pickle
+    - name: Archive artifact
+      uses: actions/upload-artifact@v2
+      with:
+          name: gw1n-9c-chipdb
+          path: apycula/GW1N-9C.pickle
+    - name: Archive stage artifact
+      uses: actions/upload-artifact@v2
+      with:
+          name: gw1n-9c-stage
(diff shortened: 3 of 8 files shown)


Makefile
```diff
@@ -5,7 +5,7 @@ endif
 .SECONDARY:
 .PHONY: all clean
 all: apycula/GW1N-1.pickle apycula/GW1N-9.pickle apycula/GW1N-4.pickle \
-	 apycula/GW1NS-2.pickle apycula/GW1NS-4.pickle
+	 apycula/GW1NS-2.pickle apycula/GW1NS-4.pickle apycula/GW1N-9C.pickle
```
yrabbit, line 8: A separate base from the GW1N-9? If the part number doesn't carry something to distinguish between these chips, then we'll have to change the nextpnr cli first, since it seems to me that it will pick up the GW1N-9 base now, right?

apycula/tiled_fuzzer.py
```diff
@@ -89,6 +89,12 @@ def recode_idx_gw1n4(idx):
         "partnumber": "GW1N-LV9PG256C6/I5",
         "recode_idx": recode_idx_gw1n9,
     },
+    "GW1N-9C": {
+        "package": "UBGA332",
+        "device": "GW1N-9C-UBGA332-6",
+        "partnumber": "GW1N-LV9UG332C6/I5",
+        "recode_idx": recode_idx_gw1n9, # TODO: recheck
```
yrabbit, line 96: Random sampling showed that the indices matched. For a total check we need a modified nextpnr.

Makefile
andelf: Yes, the cli parameter should be changed. 🥲

From the `device_info.csv`, it has two `GW1NR-LV9QN88PC6/I5`

```csv
gw1nr9-012,GW1NR-LV9QN88PC6/I5,GW1NR,GW1NR-9,QFN88P,LV,C6/I5,6,71,8640,6480,17280bit,468Kb,608Kb,20,2,4
gw1nr9c-004,GW1NR-LV9QN88PC6/I5,GW1NR,GW1NR-9C,QFN88P,LV,C6/I5,6,71,8640,6480,17280bit,468Kb,608Kb,20,2,4
```

In `gw_sh` tool, `set_device` command is used with a `-name` argument to precisely specify the device.

```tcl
set_device GW1NR-LV9QN88PC6/I5 -name GW1NR-9C
```

andelf: Actually, they are using different `IDCODE`.

andelf: The device is distinguished by the third line of silk printing, with a trailing`C`

    [2-digit-YEAR][2-digit-WEEK]C

pepijndevos: Yea the C devices tend to have their own chipdb https://github.com/YosysHQ/apicula/blob/master/doc/device_grouping.md

apycula/tiled_fuzzer.py
pepijndevos: That's good news I suppose!

pepijndevos (changes requested):
Thanks a lot for the contribution! This looks good. What's missing is an update to https://github.com/YosysHQ/apicula/blob/master/.github/workflows/chipdb.yml so that chipdbs are generated and released on CI.

It would also be great if you could update some of the examples to build for your platform by adding a constraint file and the necessary make rules. Just blinky is fine. This serves as a sanity check that we can generate bitstreams and unpack them correctly. There is no hardware in the loop but it does catch some bugs. https://github.com/YosysHQ/apicula/blob/master/examples/Makefile

Also, may I ask if you are using any publicly available board with this chip, or is it a custom design? As a matter of policy I try to test all the examples on all supported hardware, and this would be the first supported chip for which I do not have the hardware to verify its correctness.

And as @yrabbit pointed out, for full support an update to nextpnr is required.

We tend to follow the following procedure:

1. Test locally by adding it here https://github.com/YosysHQ/nextpnr/blob/fbeef2b8c2734210a70e51a3ac629c2065c06d19/gowin/CMakeLists.txt#L4
2. Merge the Apicula PR (which will be failing to build examples for this platform)
3. Do an alpha release
4. Update nextpnr here: https://github.com/YosysHQ/nextpnr/blob/fbeef2b8c2734210a70e51a3ac629c2065c06d19/.cirrus/Dockerfile.ubuntu20.04#L68
5. Do a full Apicula release, by which point CI should pass and tested on hardware

Icenowy:
@pepijndevos The board used is Sipeed Tang Nano 9K, it should be publicly available (but I heard that it runs out of stock now)

pepijndevos:
You're still missing the dependency here: `needs: [gw1n1, gw1n9, gw1n4, gw1ns2, gw1ns4]`

pepijndevos:
@Icenowy you should now be able to update nextpnr to 0.2a3 and add gw1n9c support there.

PR #78, opened 2022-01-24
chipdb: alu: mask CUPCDN functionality

Icenowy:
CUPCDN functionality seems to behave exactly the same with SUB, which
leads to unpacking failure when unpacking a bitstream that utilizes
SUB.

Mask the functionality in chipdb generation code now.

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>

### DIFF
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -198,7 +198,8 @@ def fse_luts(fse, ttyp):
             # CDN
             add_alu_mode(mode, bel.modes, lut, "7",     "0101000001011111")
             # CUPCDN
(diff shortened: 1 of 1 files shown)


pepijndevos:
Oh hehe I figured we'd update the unpacker to default to SUB if it is ambiguous, but I guess this also works for now. @yrabbit  is the one who implemented ALU for the most part so I'm not terribly familiar with that part of the unpacker code. Yosys pretty much only uses ADDSUB so not a lot will happen if we just comment out CUPCDN

PR #80, opened 2022-02-05
Add support for shadow SRAM

tpambor:


pepijndevos:
Amazing :100: 

This just extracts the main enable bit, and the rest is just normal LUT and DFF modes?

Now that we have that bit in the chipdb, the next steps are thinking about

1. unpacking
2. packing
3. nextpnr

From what I understand one `RAM16SDP4` is really a constellation of an enable bit and known primitives that span an entire tile. If you look at the `ALU` example, you can see that it copies the needed `LUT` bits to implement the full constellation. We will probably need to do the same. Except the `DFF` bits are not available at this stage, so it needs to be called after the fuzzer has run. I can write more later but have to run now.

tpambor:
> This just extracts the main enable bit, and the rest is just normal LUT and DFF modes?

The read port is async but normal DFF* can be placed in the same slice to make it synchronous. Also MUX could be used to make the read port 32x2 or 64x1. The write port seems to always be 4-bit wide.

The clock enable is always tied high. I'm not sure if it has a function in shadow RAM mode but it doesn't matter really much because it serves the same purpose as the write enable, which uses the LSR. The bit to select pos-/ or neg-edge for clock could probably be copied from DFF mode. It also seems possible to use active-low or active-high write-enable.

> If you look at the ALU example, you can see that it copies the needed LUT bits to implement the full constellation.

Yeah, that is still missing. These can be copied from LUT0-LUT3, the bits are the same.

> From what I understand one RAM16SDP4 is really a constellation of an enable bit and known primitives that span an entire tile.

I think only CLS0-CLS2 are used, so the two LUTs without DFFs are also available for use. As I mentioned above, in CLS0+1 MUX and DFF could also be used. Only LUT, MUX, DFF in CLS2 are unavailable.

pepijndevos:
Ok so this seems to successfully unpack a vendor generated bitstream. I'll go ahead and merge this very important first step.

Next we'll have to update Yosys and Nextpnr to generate bitstreams with this primitive, and then update the packer to generate them, at which point we can do a full rountrip from source to bitstream and back and test on real hardware. Exciting!

pepijndevos:
@tpambor I've started adding shadow ram in nextpnr in https://github.com/YosysHQ/nextpnr/pull/995 but it's not working yet. Any input welcome. Reading this PR maybe I need to tie the clock enable high.

uis246:
> Except the DFF bits are not available at this stage, so it needs to be called after the fuzzer has run.
This deserves a bump

pepijndevos:
Stuff changed quite a bit since this PR, I think at the moment we pack lutram in nextpnr, so in Apicula we only deal with the enable bit, nextpnr configures all the slices.

First you'd need to teach yosys about it
https://github.com/YosysHQ/yosys/blob/14d50a176d59a5eac95a57a01f9e933297251d5b/techlibs/gowin/lutrams.txt#L9

or manually instantiate it. I'm not even sure if the official vendor primitive has a property to configure this? We could add some custom property.

Xiwang is currently in the middle of rewriting the nextpnr backend, but currently there is a `pack_sram` function that would need to be taught how to create negative edge DFF when appropriate.

And then hope it actually works... unless the official IDE supports it there is no guarantee it would actually work. It's just my intuition that it reuses a lot of DFF and LUT circuitry so configuring the DFF to be negative edge MIGHT just work.

Maybe before putting in all that work it's worth just manually tweaking nextpnr output to have DFFN to see if it's supported.

PR #81, opened 2022-02-06
Initial support for GW1NZ-1

chiplet:


pepijndevos:
Exciting!

chiplet:
The fuse file parsing seems to complete successfully with these changes. However, the overall build fails in `clock_fuzzer.py` around here because it doesn't find any taps

https://github.com/YosysHQ/apicula/blob/b011a83ea0919b11e2699a477f0f8955bab6b9ad/apycula/clock_fuzzer.py#L106-L111

apycula/tiled_fuzzer.py
```diff
@@ -107,6 +107,12 @@ def recode_idx_gw1n4(idx):
         "partnumber": "GW1N-LV1LQ144C6/I5",
         "recode_idx": recode_idx_gw1n1,
     },
+    "GW1NZ-1": {
+        "package": "QFN48",
+        "device": "GW1NZ-1-QFN48-6",
+        "partnumber": "GW1NZ-LV1QN48C6/I5",
+        "recode_idx": recode_idx_gw1n1, # TODO: check
```
yrabbit, line 114: At a quick glance there are significant differences from GW1N-1, so it needs its own index recoding table. I can do it next week if you don't do it before then.

pepijndevos: Hm you think the recoding could affect the clock fuzzer? Probably not right?

pepijndevos:
```python
In [14]: f = open("/home/pepijn/bin/gowin1.9.8/IDE/share/device/GW1N-1/GW1N-1.fse", "rb")

In [15]: db = fuse_h4x.readFse(f)
check 19199191

In [16]: for row in db["header"]["grid"][61]:
    ...:     print(row)

```
GW1NZ-1:
```python
[50, 67, 67, 67, 67, 71, 66, 66, 66, 91, 66, 66, 66, 66, 66, 66, 66, 88, 89, 51]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[86, 39, 40, 41, 39, 40, 41, 80, 81, 82, 83, 84, 85, 39, 40, 41, 39, 40, 41, 87]
[126, 19, 19, 19, 19, 19, 19, 19, 19, 19, 18, 18, 18, 18, 18, 18, 18, 18, 18, 125]
[64, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 65]
[64, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 65]
[64, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 17, 65]
[48, 70, 70, 70, 70, 70, 70, 70, 70, 92, 70, 70, 70, 70, 70, 70, 70, 70, 70, 49]
```
GW1N-1:
```python
[50, 66, 66, 66, 66, 66, 66, 66, 66, 91, 66, 66, 66, 66, 66, 66, 66, 88, 89, 51]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[86, 39, 40, 41, 39, 40, 41, 80, 81, 82, 83, 84, 85, 39, 40, 41, 39, 40, 41, 87]
[55, 14, 14, 14, 14, 14, 14, 14, 14, 14, 13, 13, 13, 13, 13, 13, 13, 13, 13, 54]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[64, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 65]
[48, 63, 63, 63, 63, 63, 63, 63, 63, 92, 63, 63, 63, 63, 63, 63, 63, 63, 63, 49]
```
So seems like we have some new and exciting tile types. The lower half now seems to have shadow ram, which is great. But more relevant to this issue, is that the row with the taps has tile types 18 and 19. So I believe changing the line you found to the following will at least make some progress. But no doubt there could be more surprises down the road.
```python
taps = [r for (r, c, typ), t in sweep_tiles.items() if typ in {13, 14, 15, 16, 18, 19}] 
```

apycula/tiled_fuzzer.py
yrabbit: Clock fuzzer? Oh, I don't think so. I'm only looking at I/O. The clock fuzzer should not be affected in any way.

pepijndevos: Yea I looked at the tile grid and I think I know what the problem is with the clock fuzzer, pushed a fix, so we'll know soon enough.

yrabbit: This is an example of a theoretical recoding based on the numbers in the tables. Requires experimental verification to be sure, of course. 
```
def recode_idx_gw1nz_1(idx):
    new_idx = idx
    if idx >= 40:
        new_idx -= 10
    if idx >= 60:
        new_idx -= 2
    if idx >= 80:
        new_idx -= 6
    return new_idx
```

pepijndevos: @chiplet are you able to test this? It'll be a while before I have a new Tang Nano 1K.

pepijndevos: Here is an example of adding a new chip on the nextpnr side: https://github.com/YosysHQ/nextpnr/pull/909/files

chiplet: At least the blinky seems to work now! Does this mean that the flow works correctly for all IO or should I manually verify each pin as well?

pepijndevos:
I think you may have forgotten to commit `tangnano1k.cst`?

blueskull:
A little bug was found in gowin_pack.py. The tool only accepts -d GW1NZ-1, not -d GW1NZ-LV1QN48C6/I5.

The suspected culprit is line 344:

`m = re.match("GW1N(S?)[A-Z]*-(LV|UV|UX)([0-9])C?([A-Z]{2}[0-9]+P?)(C[0-9]/I[0-9])", device)`

My proposed fix is:

`m = re.match("GW1N(S|Z)?[A-Z]*-(LV|UV|UX)([0-9])C?([A-Z]{2}[0-9]+P?)(C[0-9]/I[0-9])", device)`

Other than that, I found this new device support works wonders. Thanks for all of you for the great works!

chiplet:
@blueskull Nice find! This PR has already been merged, so you could open a new one for your patch.

PR #82, opened 2022-02-16
Switch to read the banks encoding from .FSE

yrabbit:
- store which mode is used for the case of inputs only;
  - handle multiple banks in the same corner tile;
  - add clock routes to the routing bits;
  - warn if wrong IOSTD used for pin;
  - use strings as the bank indices;
  - the new IDE does not use a device configuration file, so do not write it;
  - recoding update for GW1N-9;
  - LVCMOS12 is back;
  - document the 37 longval table.

### DIFF
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -46,6 +46,7 @@ class Bel:
     # banks
     bank_mask: Set[Coord] = field(default_factory=set)
     bank_flags: Dict[str, Set[Coord]] = field(default_factory=dict)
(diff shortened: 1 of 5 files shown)


yrabbit:
the bases before GW1NZ-1 should be correct. GW1NZ-1 bases are mostly correct, possible effects when using PCI33, I will deal with this later.
Now the relevant is the next step --- pure differential output without logic.

pepijndevos:
I'll test this PR next weekend probably

PR #84, opened 2022-02-17
These cells, found in particular in GW1NZ-1, also have DFFs.

yrabbit:


pepijndevos:
thanks :)

PR #86, opened 2022-02-23
Add support for true differential output (TLVDS)

yrabbit:
Also:
- extraction of information about the differential signal support in the pins;
- changed the search system in longval tables


pepijndevos (commented):
looks fine I guess.

Would be cool if we had some example to test this.

yrabbit:
I agree. For tangnano 4k we don't even have a blinky:)

yrabbit:
The simplest option would require an oscilloscope to confirm the polarity reversal of the signal. Do you think it's worth including one of these? 
It is possible, of course, to check with a pair of LEDs turned on in the opposite direction to each other... hmmm... a bit tricky.

```
module top (
    input clk,
    output o_led_p,
    output o_led_n
);

reg [25:0] ctr_q;
wire [25:0] ctr_d;
wire i_led;

// Sequential code (flip-flop)
always @(posedge clk)
    ctr_q <= ctr_d;

// Combinational code (boolean logic)
assign ctr_d = ctr_q + 1'b1;
assign i_led = ctr_q[25:25];

TLVDS_OBUF led(
        .O(o_led_p),
        .OB(o_led_n),
        .I(i_led)
    );

endmodule
```

pepijndevos:
The [tonegen](https://github.com/YosysHQ/apicula/tree/master/examples/tonegen) demo already requires a scope to test so this is absolutely fine.

yrabbit:
Done.
Although I desperately want 135MHz, I even read 1 page about PLLVR, didn't understand anything :)

yrabbit:
All boards except NZ-1 and N-1 have these pins, but I do not have a pinout, so I just do not know the physical availability of the pins, although I know which pins of the chip itself can be used.

examples/Makefile
```diff
@@ -4,7 +4,8 @@ NEXTPNR ?= nextpnr-gowin
 all: attosoc-tec0117.fs nanolcd-tangnano.fs blinky-tec0117.fs blinky-runber.fs \
 	blinky-tangnano.fs blinky-honeycomb.fs shift-tec0117.fs shift-runber.fs \
 	shift-tangnano.fs shift-honeycomb.fs iob-tangnano.fs iob-default-flags-tangnano.fs \
-	tonegen-tec0117.fs blinky-tangnano9k.fs blinky-tangnano1k.fs
+	tonegen-tec0117.fs blinky-tangnano9k.fs blinky-tangnano1k.fs \
+	tlvds-tangnano4k.fs tlvds-tangnano9k.fs tlvds-tec0117.fs
```
pepijndevos, line 8: should also be added to the unpack target (did a bunch of those in #87)

pepijndevos:
The generated examples work for me on tang nano 4K but do nothing on TEC0117. That is, one output is constantly low while the other is constantly high.  I wonder if the actual LVDS is broken or the counter logic is bugged leading to a constant output.

Tang nano 9K is in the country so can test that one soon.

yrabbit:
I should note that all True LVDS pins on the tangnano4k board on the master branch are used for LEDs in the examples so here I take away two pins and reduce the number of available LEDs (not that it matters since they are all external).

Issue #85, opened 2022-02-20
Need to update the provided exemple in readme.md


Popolon:
In the readme.md of apicula:

https://github.com/YosysHQ/apicula/blob/master/readme.md
<pre>yosys -D LEDS_NR=8 -p "synth_gowin -json blinky.json" blinky.v</pre>

This works fine with yosys 0.12. since  yosys 0.13, yosys need to have the read_verilog instruction to work and the verilog files are passed to it instead of the end of the line:
<pre>yosys -D LEDS_NR=8 -p "read_verilog blinky.v; synth_gowin -json blinky.json"</pre>

Both syntax depending on version should so be described.

yrabbit:
It would be just great if you could form it as PR. I would be happy to merge it ;)

pepijndevos:
This was fixed in #88 I believe

PR #88, opened 2022-02-28
Improvements to readme example

DrJosh9000:
See #85 re the invocation of Yosys. This also uses some variables instead of requiring changing device/board in multiple places.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -40,15 +40,17 @@ You can also use the Makefile in the examples folder to build the examples.
 
 ```bash
 cd examples
-yosys -D LEDS_NR=8 -p "synth_gowin -json blinky.json" blinky.v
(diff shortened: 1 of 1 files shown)


pepijndevos:
Don't those variables need `export` to work when pasted in a shell?

trabucayre:
Those variables are used in current shell, so there is no need to use `export`. This keyword is only required when used in subcommands

uis246:
`export` should be passed.

1. `A=b`
2. `env | grep A=`
Result: no A

1. `export A=b`
2. `env | grep A=`
Result: A=b

trabucayre:
yes but
````
$ TOTO="a"
$ echo $TOTO
a
```
here variable are directly used without needs of environment. But using `export` is not bad at all ;-)

pepijndevos:
Funny I never picked up on the subprocess environment distinction. Makes sense!

PR #89, opened 2022-03-12
Add the Global Set/Reset primitive

yrabbit:
It's actually empty, but it's the routing that counts.


yrabbit:
Hmm... Let's slow down for a while. It seems that if the VCC is very far away from the cell (0, 0), the chip may not start. Clearly, in the vendor version, the VCC is right there in the same cell.

It is not worth merging for now. 

There will be no problem with nextpnr, it will not generate this primitive if the bases do not support it.

PR #90, opened 2022-03-13
VCC to GSR hack

yrabbit:
Connect the VCC to the GSR input directly in the template and remove the
ability to use these pips for routing.


yrabbit:
So far I have only checked for the GW1N-9C.

pepijndevos (approved):

PR #93, opened 2022-03-23
Change regex pattern to better support GW1NZ with full part name

blueskull:


pepijndevos:
Good catch! Thanks :+1:

PR #95, opened 2022-03-25
Connect the outermost cells to the clock.

yrabbit:
Extrapolation of the clock wires to the outermost cells. This is
essentially an old fuzzer with a presumed location of the clock wires.
The probability of misplacement exists, but it is small.

### DIFF
--- a/apycula/clock_fuzzer.py
+++ b/apycula/clock_fuzzer.py
@@ -275,7 +275,7 @@ def add_rim(rows, cols):
         rows.update({row for row in range(max(rows) + 1, db.rows)})
     if 1 in cols:
         cols.add(0)
-    else:
+    elif db.cols - 2 in cols:
         cols.add(db.cols - 1)
     return rows, cols
 
--- a/apycula/clock_fuzzer.py
+++ b/apycula/clock_fuzzer.py
@@ -266,9 +266,23 @@ def spine_aliases(quads, dests, clks):
                     aliases[spine_row, tap, dest] = ct[0], ct[1], dest
     return aliases
 
+# add border cells
+def add_rim(rows, cols):
+    if 1 in rows:
+        rows.add(0)
+    else:
+        # XXX fill all rows
+        rows.update({row for row in range(max(rows) + 1, db.rows)})
+    if 1 in cols:
+        cols.add(0)
+    else:
+        cols.add(db.cols - 1)
+    return rows, cols
+
 def tap_aliases(quads):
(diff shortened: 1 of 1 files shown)


apycula/clock_fuzzer.py
```diff
@@ -266,9 +266,23 @@ def spine_aliases(quads, dests, clks):
                     aliases[spine_row, tap, dest] = ct[0], ct[1], dest
     return aliases
 
+# add border cells
+def add_rim(rows, cols):
+    if 1 in rows:
+        rows.add(0)
+    else:
+        # XXX fill all rows
+        rows.update({row for row in range(max(rows) + 1, db.rows)})
```
pepijndevos, line 275: What is going on here? Don't have the datastructure fresh in my mind

apycula/clock_fuzzer.py
```diff
@@ -266,9 +266,23 @@ def spine_aliases(quads, dests, clks):
                     aliases[spine_row, tap, dest] = ct[0], ct[1], dest
     return aliases
 
+# add border cells
+def add_rim(rows, cols):
+    if 1 in rows:
```
pepijndevos, line 271: so does this only deal with the left rim? What about the `db.rows` column/row? I'd assume we need to add those to the next one as well.

apycula/clock_fuzzer.py
yrabbit: When you define row and column ranges for quadrants, the rows that do not contain any DFF do not get into the list of quadrant rows, they are rows with memory apparently. Here I still add these rows so that then they have aliases.

yrabbit: option is not very good because it adds only the outermost lines, but I do not have a good idea where the memory is usually located, I assume that at the edge of the quadrant.

apycula/clock_fuzzer.py
yrabbit: they are added via else: clauses

apycula/clock_fuzzer.py
```diff
@@ -286,6 +300,7 @@ def branch_aliases(quads, clks):
             else:
                 src = "GBO1"
             for tap, branch_cols in taps.items():
+                add_rim(rows, branch_cols)
```
pepijndevos, line 303: So the else clause adds the last column to every single branch? That seems incorrect.

yrabbit: you're right

PR #96, opened 2022-03-26
Add support for the ODDR primitive

yrabbit:
As well as examples of using ODDR with or without TLVDS.


pepijndevos:
Confirmed working on tec0117

PR #107, opened 2022-05-28
Add the long wires

yrabbit:
It becomes possible to use fast wires that go into each cell of the
chip. To do this, just mark the wire in the .CST file as follows:

CLOCK_LOC "auto_net" BUFS;
CLOCK_LOC "net_with_number" BUFS[6];

the first option is used when the wire number used is indifferent (and
there are only 8 of them), the second option specifies a number.

One can use long wires to OBUF, but other IO primitives will probably
switch to normal routing for now.

See doc/longwires.md


yrabbit:
What works: 
On tangnano, tangnano4k, tangnano9k, GW1NR-9 (szpga):
shift, blinky, blinky-oddr, tlvds, tlvds-oddr

Need to check on GW1NS-2, GW1N-4, GW1N-9(tec0117), GW1NZ-1 (I don't have these boards)

What doesn't work is blinky-osc-tangnano and blinky-osc-tangnano4k, but I should note that they didn't work in https://github.com/YosysHQ/apicula/actions/runs/2071947590#artifacts either

pepijndevos:
Hmmmm what happened to the oscillator examples... :thinking: 

Of course happy to test some bitstreams on my boards.

yrabbit:
> Of course happy to test some bitstreams on my boards.

I just realized that I didn't mention here that the test bitstreams are all examples because they specify to use long wires for LEDs:)
So would appreciate running blinky on the boards I don't have.

yrabbit:
Or not blinky --- as I said all the examples use long wires. True, I don't expect attosoc to work, but that's due to our other obscure fiddlesticks:)

pepijndevos:
Seems to work on tec0117, runber, and honeycomb! So lmk when this is ready to merge

yrabbit:
Yey!
It remains to decide whether or not to remove the total use of long wires from the examples.

apycula/tiled_fuzzer.py
```diff
@@ -101,45 +101,174 @@ def recode_idx_gw1nz_1(idx):
...
+        "has_bottom_quadrants": True,
+        "lw_taps": [3, 2, 1, 0],
     },
     "GW1N-9C": {
         "package": "UBGA332",
         "device": "GW1N-9C-UBGA332-6",
         "partnumber": "GW1N-LV9UG332C6/I5",
-        "recode_idx": recode_idx_gw1n9, # TODO: recheck
+        "recode_idx": recode_idx_gw1n9,
+        "has_bottom_quadrants": True,
```
pepijndevos, line 128: the clock fuzzer already figures this out without hardcoding values, maybe we can reuse that?

yrabbit: Yes, that would be correct. Since the clock fuzzer hasn't worked yet at this point, I'll probably move the whole thing to clock_fuzzer.py.

yrabbit: After all, long wires are also a kind of clock.
By the way, a side effect is that there is a free gate from logic to clocks: the input points of the long wires can be used by nextpnr as sources for clock SPINExx

For good or bad, we will see:)

yrabbit: Now we use the data from the clock fuzzer! Unfortunately GW1NZ-1 is not friendly and clock 3 is not found:(

pepijndevos: :(

yrabbit:
In order not to touch yet the clock GW1NZ-1 we fill in the missing value of the tap column of long wires.

pepijndevos:
Looks good?

Thinking about it a bit more, I think I actually prefer a separate long-wire blinky. For debugging stuff you usually want stuff to be as simple as possible.

yrabbit:
Okay, I'll redo the examples. I think I'll throw out the *-default-flags* stuff as well: it was useful a year ago when IO flags were under research, but now it just wastes compile time.

pepijndevos:
hurray!

PR #121, opened 2022-07-22
packer: generalization of ignored cell types

yrabbit:
* Experiments with specialized routers can lead (in intermediate
  branches they already do) to the use of fake cell types, which are not
  encoded in any way, but only used to create the desired sequence of
  PIPs. To avoid adding a lot of ignored types, it is proposed to ignore
  types starting with DUMMY_.
* Removed swearing on the unknown type MUX2_, but the type itself is
  left for the reason that although no action is required from the
  packer, it is still a real cell type.


pepijndevos:
Oh the CI still fails with the RAM error I should really really fix that... Other than that it seems fine.

PR #124, opened 2022-08-13
Update readme.md

Joshinken:
took me way too long to realize that im not pepijn. decided to make it slightly clearer to save anyone else the embarassment


trabucayre:
Path you mention is always at the same location: maybe simply using `$HOME` instead `/home/YOURUSERNAME` will be better since this variable is always defined, and is the absolute path to this directory.

uis246:
export PATH="$HOME/.local/bin:$PATH"

pepijndevos:
Closing this and just updating the readme myself with the suggested `$HOME` variable

Joshinken:
I forgot this was even a thing

PR #125, opened 2022-08-16
Fix blinky for tangnano4k

uis246:
And also rename hdmi pins

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -120,7 +120,7 @@ nanolcd-tangnano-synth.json: nanolcd/TOP.v nanolcd/VGAMod.v
 	$(YOSYS) -D LEDS_NR=3 -D OSC_TYPE_OSCZ -p "read_verilog $^; synth_gowin -json $@"
 
 %-tangnano4k-synth.json: %.v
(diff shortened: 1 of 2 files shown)


yrabbit:
Ignoring the LED on this board is not a mistake, but a precautionary measure: any failure when using the MODE pin as a GPIO can turn your FPGA into a brick, and on this board the LED is just on the MODE pin (10).
The ability to use the MODE pin as a GPIO is not even supported by gowin_pack.
So this board requires external LEDs to be connected, well since we are connecting external LEDs anyway, why 1 and not 6? :)
I don't think renaming these lines is a good idea - it would break some examples.

uis246:
But my board wasn't bricked? Only write to flash bricks?

yrabbit:
> But my board didn't brick. Only write to flash bricks?
> 
> > The ability to use the MODE pin as a GPIO is not even supported by gowin_pack.
> 
> IDK. Worked for me.
Now that's a real mistake :) We should take a look at the code and fix it.

pepijndevos:
Feel free to reopen if this is still relevant

PR #126, opened 2022-08-18
Add a check on the use of dangerous pins

yrabbit:
gowin_pack terminates if it detects the use of dangerous pins as GPIOs.
There is no command line key that allows the use: it is assumed that in
case of emergency one can make changes in gowin_pack, being
fully aware of the consequences.

### DIFF
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -80,6 +80,7 @@ class Device:
     packages: Dict[str, Tuple[str, str, str]] = field(default_factory=dict)
     pinout: Dict[str, Dict[str, Dict[str, str]]] = field(default_factory=dict)
     pin_bank: Dict[str, int] = field(default_factory = dict)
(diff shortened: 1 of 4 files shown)


pepijndevos:
haha editing the code seems maybe a bit drastic. Given that some boards just use them as normal IOs it's definitely something we should consider supporting better in the future. In the vendor IDE they are just checkboxes you can enable, so having them as command line flags seems appropriate.

Since @uis246 used the mode pin without bricking the FPGA it's at least not completely broken...

Maybe I should refresh my memory on what their functions and dangers are and how to recover from them, do some testing, and then just support these pins?

yrabbit:
Makes sense, I based it on your experience:)
In the current state, the MODE pins are no different from the normal pins. We can leave them that way. 
To this simple fix I have no special attachment:)

PR #128, opened 2022-09-16
Redo to autodetect the location of I/O banks

yrabbit:


yrabbit:
Well, I think that nextpnr-0.3 is already too old and will always swear:)

pepijndevos:
Whoa what is this magic, "TODO" comments actually getting fixed

PR #129, opened 2022-10-16
Place logicinfo in the database.

yrabbit:
The logicinfo table describes the valid values of the parameters of the primitives (bels). Each line is a pair (parameter_id, value). This table can be used to check if the parameter values are set correctly, but its most important purpose is to generate fuses that program the above parameter value.  The fact is that the first two fields of the shortval table are the row numbers of this logicinfo table, that is, the fuses that are listed in the row of the shortval table must be set to set the values of this parameter pair.

As long as there is no decoding of parameter_id it is hard to give a clear example.  From unconfirmed theory in the longval table, the first six fields also refer to logicinfo.


yrabbit:
ugh, finding a good example with 'shortval' is difficult: even flip-flops are constructed of several parts like "turn off/off inverter", "set MUX for clk", "set MUX for CE" etc. I have no idea. 
I'll have to research 'longval' because these tables are used for IO, and I feel much more confident there:)

yrabbit:
Great! The heart of the system, the function to pull fuses by attributes, is in place. Now it's up to the attributes themselves and their values.

yrabbit:
Well it turns out that the table 'logicinfo' is not suitable for a full-fledged check of valid values - it does not store default values.
This is inconvenient:(

yrabbit:
This is the last of the auxiliary functions needed to get the fuse in the "new", correct style.
Next, it is a matter of reworking the existing coding parts. 

And the basic idea is to manipulate and perform checks at a higher level than the bits of a particular chip - that is, to combine the values of the attributes of the primitive and then get the bits for a particular chip in one move.

This is the ideal. In practice, I foresee there will still be checks on the current chip, but it will be about attribute values, not bits.

PR #132, opened 2022-11-09
Make less dependencies required

mmicko:
To make apicula available in oss-cad-suite build I needed to make sure all python libraries are delivered as well, and there is problem building `pillow` in cross compile environment, also `pandas` was not used at all (and was causing some problems when doing cross compilation as well). 
Since `-png` option is anyway only for development, I just made sure it is available only if `pillow` is installed.

### DIFF
--- a/.github/workflows/chipdb.yml
+++ b/.github/workflows/chipdb.yml
@@ -178,7 +178,7 @@ jobs:
         name: gw1ns-4-chipdb
         path: apycula
     - name: Set up Python
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
       with:
         python-version: '3.8'
     - name: Install and build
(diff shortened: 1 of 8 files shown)


pepijndevos:
Great! Actually I don't think we need openpyxl either, it was used to parse excel sheets which we're no longer doing.

pepijndevos:
Thanks!

pepijndevos:
@mmicko could this be the result of your change? https://github.com/YosysHQ/apicula/releases/tag/0.0.0.dev

mmicko:
Hi. Sorry for not pinging you first. But it is neceserry to have tag so I
can have release that I will update with latest pickle files. So please do
not delete this one.

Oss-cad-suite scripts are made and already using this. Please do try
apicula using oss-cad-suite-build :)

pepijndevos:
So will each release update this dev release? Why not attach it to the new release? If it's expected then all good, was just confused.

PR #133, opened 2022-11-10
Add initial PLL support

yrabbit:
This PLL only works on the GW1N-1 chip and understands only two fixed frequencies: 52 and 56 MHz.

Dynamic divider tuning is not supported.


yrabbit:
There is no example and no unpacker. And PR also fully includes logicinfo, just without the lengthy comments on commits :)

yrabbit:
I think the unpacker is not wrong in defining PLL parameters in vendor images, but is wrong with CLKIN and CLKFB because of the two hidden MUX on these inputs.

We will deal with them later when special wires for such clock signals are investigated and activated.

chili-chips-ba:
How is the current level of on-chip PLL support?
The question is mostly for the device onboard TangNano20K...

yrabbit:
Of the 4 PLLs on Tangnano20k, the bottom two are supported. The top pair is not working and is prohibited even in the vendor's IDE.

Please note that Tangnano20k is only supported in the Himbaechel version, not in the legacy gowin.

PR #134, opened 2022-11-15
Add parameter check and DUTY handling

yrabbit:
As stated in the documentation: "For example, when the phase shift setting is "0" (00000), 50% duty cycle will be "8" (1000). If the phase shift setting is "180", the 50%, the 50% duty cycle setting is "0" (0000)."

As there are fuses for all DUTY values at the moment we do not check for special missing cases 0, 1, 15.

In the case of static divider assignment, check the permissible frequency range for the desired chip.


yrabbit:
I think the PLL examples need to be reworked so that several images with different output frequencies are generated for each board. hm...

yrabbit:
About support for different feedback and clkin routes - gowin_pack of course does not do routing, it just removes restrictions on usage, the routes themselves are created in nextpnr :)

PR #135, opened 2022-11-25
Add information about the functions of the pins.o

yrabbit:
Includes information on additional pin functions such as RPLL_C_IN, GCLKC_3, SCLK and others.
This allows a decision to be made about special network routing of such pins


yrabbit:
Unfortunately the version of the chip base is changing and this will entail an interesting process of friendship with nextpnr.

pepijndevos:
`ERROR: Incorrect chipdb version 2 is used. Version 1 is required` yea

PR #139, opened 2022-12-13
Add ability to use the clock net for PLL signals

yrabbit:
The nextpnr can now use the clock network to deliver PLL outputs if it sees fit.

On the apicula chip base side, all four outputs can be connected to the clock network, but the nextpnr currently has a limit of 3 networks, if I'm not mistaken. This is not being touched for the time being.

In the process of experimentation, the functions of the many internal wires of the large central clock MUX have been discovered and these are reflected in the names.

An example of using the PLL to form a picture on a 4.3" LCD screen is added.

! This commit uses the existing clock network processing logic and does not require any changes to the nextpnr, i.e. no new information about the big MUX device will be applied.


pepijndevos:
Amazing! Looks good and confirmed it works on my end.

PR #146, opened 2022-12-27
 Don't create GTx aliases along the central column

yrabbit:
These cells are served by the closest tap on the west side (according to GBx0 clock wire number) and clock fuzzer finds this perfectly, no GTx0 wires are used in the central column.

This did not cause errors, just no need to create wires leading nowhere and waste routing time.


yrabbit:
Oops, I grabbed an extra

PR #148, opened 2022-12-30
Do not send nextpnr wires of dubious purpose

yrabbit:
The presence of these PIPs does not mean that they are functional, so until their mechanisms are clarified, we do not use them.


yrabbit:
I made the cutoff in gowin_bba because the unpacker is too useful and on the apicula side it is better to have complete bases, it is only nextpnr not necessarily aware of questionable routes:)

PR #149, opened 2023-01-02
Bring examples into line with the current config

yrabbit:
Now that the global router is enabled by default, there is no need for separate examples for it.
Also, long wires are disabled (but not removed) because undocumented PIPs are used there.


yrabbit:
The long wires are the local bus network I'm dealing with at the moment, but I don't think it's wise to divert my attention to fixing the examples just yet;)

PR #150, opened 2023-01-08
Add the alonenode mechanism

yrabbit:
Storing and using part of the "alonenode" table to set fuses designed to disable "lone" nodes, i.e. nodes that are not used as sources. It is used to switch long wires in buses.

In addition, the representation of long wires is greatly simplified by creating only the necessary aliases and transferring the main work to the nextpnr.

It should be noted that only table #6 of the "alonenode" family is involved, since the function of table #69 remains unclear.


pepijndevos:
Removing a bunch of fuzzing is exciting progress!

PR #151, opened 2023-01-11
GW1N-9x: Add two rows as normal clock ones

yrabbit:
These rows are also nothing special, there are working GB and GT pips.


yrabbit:
It still doesn't fix attosoc, but step by step the suspicious places are cleaned up. :)

PR #154, opened 2023-01-19
stage 2. Add GW1NSR-4C PLLs

yrabbit:
A new type of PLL is added - PLLVR, which is used in the GW1NS-4 chips. Both PLL instances are packaged and unpacked with support for all attributes, including static and dynamic setting of generated frequency parameters.

Describes the implicit wires leading to the central clock MUX, which allows us to use all 4 PLL outputs using the global clock networks.

Also the meaning of Table 59 and a couple of attributes in it was discovered. Not used yet, the constants added for the future.


uis246:
Thank you

pepijndevos:
Kinda ran out of weekend to do any testing on this but CI is green, it looks fine and I saw it working so I'll just merge it, and maybe do some tests before the next release.

Issue #158, opened 2023-01-30
gowin_pack fails on nano9k with strange LVCMOS18 error


martell:
I've been trying to build my_hdmi_device by splinedrive with apicula and have gotten pretty far along with some tweaks.
gowin_pack with the current apicula/chipdb fails with this strange error on the nano9k

```
gowin_pack -d GW1N-9C -o chip_balls.fs chip_balls_pnr.json
Traceback (most recent call last):
  File "/opt/homebrew/bin/gowin_pack", line 8, in <module>
    sys.exit(main())
  File "/opt/homebrew/lib/python3.10/site-packages/apycula/gowin_pack.py", line 703, in main
    place(db, tilemap, bels, cst, args)
  File "/opt/homebrew/lib/python3.10/site-packages/apycula/gowin_pack.py", line 452, in place
    raise Exception(f"Different I/O modes for the same bank {bank} were specified: {iostd} and {flag_name_val[1]}")
Exception: Different I/O modes for the same bank 1 were specified: LVCMOS18 and LVCMOS33
make: *** [chip_balls.fs] Error 1
```

Steps to reproduce

```
git clone https://github.com/martell/my_hdmi_device
cd my_hdmi_device
git checkout yosys-nano9k
make -f Makefile.nano9k
```

@yrabbit @pepijndevos This seems very close since merging #156

yrabbit:
Let's see if it doesn't get resolved after we beat 

https://github.com/YosysHQ/apicula/issues/157

martell:
Continuing from there trying to use LVDS here this is where we get stuck

I thought the nano9k had emulated lvds instead of true from reading docs and this reddit thread proved useful.
https://www.reddit.com/r/GowinFPGA/comments/wu96b5/difference_between_gowin_true_and_emulated_lvds/

Using ELVDS_OBUF I get the following error - I presume apicula just doesn't have support for it yet ?
```
7.3.21. Executing AST frontend in derive mode using pre-parsed AST for module `\ball'.
Generating RTLIL representation for module `$paramod$df58ad4b67c0055eee254f8d5891cca82e58fb35\ball'.
ERROR: Module `\ELVDS_OBUF' referenced in module `\chip_balls' in cell `\clk_tlvds_obuf' is not part of the design.
```

Using TLVDS gowin pack errors out with this
```
Exception: IOT42A. TLVDS_OBUF is not allowed for a given io standard LVCMOS33
```

[...]

PR #159, opened 2023-01-30
Add support for emulated differential output (ELVDS)

martell:
This is a draft - I still need to figure out what I am doing 

Will use https://github.com/YosysHQ/apicula/issues/158#issuecomment-1407959837
and https://github.com/YosysHQ/apicula/pull/86 as a guide.

Thanks again @yrabbit 

EDIT: looks like the nano9k still has TLVDS so need to keep that for that board still

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3
+FROM --platform=linux/amd64 python:3
 
 WORKDIR /usr/src/gowin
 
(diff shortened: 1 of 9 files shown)


martell:
I added a PR to yosys here https://github.com/YosysHQ/yosys/pull/3652

pepijndevos:
Wow I look away for a second and we have a cool new feature and cool new contributor!

I see a lot of stuff in my mailbox, just in case, we also have a matrix/irc channel for more real time discussion.

yrabbit:
The developments are taken into account in the PR https://github.com/YosysHQ/apicula/pull/177

PR #161, opened 2023-02-04
stage 2. Add PLL support for GW1NR-4 chips

yrabbit:
These are interesting chips, the two PLLs are asymmetrically placed in one half of the chip and also occupy only two cell


uis246:
Unstopable rabbit

PR #164, opened 2023-02-06
Replace DFF fuzzer with table data

yrabbit:
DFFs are assembled from parts depending on the type, also when unpacking the type is reconstructed from the used functional parts, which means "clock line inversion", "asynchrony", "triggering by SET or RESET", etc.

### DIFF
--- a/apycula/gowin_pack.py
+++ b/apycula/gowin_pack.py
@@ -13,7 +13,7 @@ from apycula import codegen
 from apycula import chipdb
 from apycula.chipdb import add_attr_val, get_shortval_fuses, get_longval_fuses
 from apycula import attrids
-from apycula.attrids import pll_attrids, pll_attrvals
+from apycula.attrids import pll_attrids, pll_attrvals, cls_attrids, cls_attrvals
 from apycula import bslib
 from apycula import attrids
 from apycula.wirenames import wirenames, wirenumbers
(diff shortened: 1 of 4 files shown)


apycula/gowin_pack.py
```diff
@@ -392,11 +392,31 @@ def place(db, tilemap, bels, cst, args):
...
+                if mode in {'DFFR', 'DFFC', 'DFFNR', 'DFFNC'}:
+                    add_attr_val(db, 'SLICE', dff_attrs, cls_attrids[f'REG{int(num) % 2}_REGSET'], cls_attrvals['RESET'])
+                elif mode not in {'DFF', 'DFFN'}:
+                    add_attr_val(db, 'SLICE', dff_attrs, cls_attrids[f'REG{int(num) % 2}_REGSET'], cls_attrvals['SET'])
+                # are set/reset/clear/preset port needed?
+                if mode not in {'DFF', 'DFFN'}:
+                    add_attr_val(db, 'SLICE', dff_attrs, cls_attrids['LSRONMUX'], cls_attrvals['LSRMUX'])
+                # invert clock?
+                if mode in {'DFFN', 'DFFNR', 'DFFNC', 'DFFNP', 'DFFNS'}:
+                    add_attr_val(db, 'SLICE', dff_attrs, cls_attrids['CLKMUX_CLK'], cls_attrvals['INV'])
```
uis246, line 409: Possibly you need here
```
else:
add_attr_val(db, 'SLICE', dff_attrs, cls_attrids['CLKMUX_CLK'], cls_attrvals['SIG'])
```

yrabbit: Common sense dictates that the default values should not have fuses set, but in case we come across such a chip, so we add it.

Thanks;)

PR #170, opened 2023-03-23
Add support for OSER primitives

yrabbit:
Primitives are implemented that convert the parallel signals into a serial signal. The primitives differ in the number of input parallel signals: 4, 7, 8 and 10 respectively for OSER4, OVIDEO, OSER8 and OSER10 primitives.

The serialization primitives require two clock frequencies: one slow (pclk) is used to load parallel data, and the other fast (fclk) is used to generate the serial output signal.

For pclk, normal CLK-type ports are used, which under certain conditions can use routing through global clock networks or ordinary wires - nothing unusual here.

But fclk uses a special kind of wires - HCLK, of which there are very few: 4 wires per side of the chip, and you have to choose in each particular IO buffer from only two wires.

Not all chips have all sides covered by HCLK: the GW1N-1 has only the bottom side, the GW1NZ-1 has only the top and right side, the GW1NSR-4C has all but the left side, etc.

The situation is not improved by the fact that these wires are controlled in special cells, the nature and number of which depend on the chip.

This commit implements the HCLK capabilities necessary for the operation of the serialization primitives.

The OSER4, OVIDEO, OSER8 and OSER10 primitives are implemented for the GW1N-1, GW1NZ-1, GW1N-4, GW1NSR-4C, GW1NR-9 and GW1NR-9C chips.

Demo examples are added for all primitives. On the tangnano9k board you can observe the operation immediately on the LEDs (because there are 6 of them), for other boards you will need a logic analyzer.


yrabbit:
Running all the examples is somewhat tedious, and in the case of IDESx also requires external clock and signal sources. As a quick and convenient option I can advise to connect Tangnano9k and watch funny flickering on its 6 LEDs :)

One LED is a slow frequency, the second is a fast frequency, the rest are outputs of individual OSERx primitives.

#this one stops on pressing the button
openFPGALoader -c ft2232 examples/blinky-oddr-tangnano9k.fs      

# PCLK = 1/2 FCLK
openFPGALoader -c ft2232 examples/deser-oser4-tangnano9k.fs 

# PCLK = 1/3.5 FCLK  it's fun :)
openFPGALoader -c ft2232 examples/deser-ovideo-tangnano9k.fs 

# PCLK = 1/4 FCLK
openFPGALoader -c ft2232 examples/deser-oser8-tangnano9k.fs

# PCLK = 1/5 FCLK
openFPGALoader -c ft2232 examples/deser-oser10-tangnano9k.fs

OSER16 can not be output to LEDs because the chip side with LEDs does not have such primitives.

pepijndevos:
whoa that's a lot of examples and big changes but all the stuff I looked at and tested looks reasonable haha amazing stuff

PR #174, opened 2023-04-20
Added new oscillator types and improved its handling

yrabbit:
Revised PR, originally by uis246

* Take into account the fact that some inputs and outputs are in other cells from the primitive
* Clear energy conservation bits inherited from image template
* Enable OSC in the GW1NZ-1 chip by setting the impirical bits.
* Minor edits.


yrabbit:
Works on tangnano, tangnano1k, tangnano4k and tangnano9k

PR #178, opened 2023-06-09
adding little helper tool like icepll

multigcs:
written in python,
only for the GW1N-9C
finds the best match

### DIFF
--- a/gowin-pll.py
+++ b/gowin-pll.py
@@ -1,0 +1,99 @@
+#!/usr/bin/python3
+#
+# pll tool to find best match for the target frequency (GW1NR-9C C6/I5 (Tang Nano 9K proto dev board))
+
+
+import argparse
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "-i", "--input-freq-mhz", help="PLL Input Frequency", type=float, default=27
+)
+parser.add_argument(
+    "-o", "--output-freq-mhz", help="PLL Output Frequency", type=float, default=108
+)
+
+parser.add_argument(
+    "-f",
+    "--filename",
+    help="Save PLL configuration as Verilog to file",
+    type=str,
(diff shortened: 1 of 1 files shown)


gowin-pll.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/python3
```
yrabbit, line 1: ` #!/usr/bin/env python3`

?

yrabbit:
Great stuff! It generates working code for 9C.
How about adding a board selection parameter like `-d GW1NZ-1` ?

multigcs:
i have added more chip types but i can not test this all, i have only the TangNano9K board

yrabbit:
Great job! I will play with different boards in the next day or two.

yrabbit:
Works for Tangnano4k and Tangnano9k

gowin-pll.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "-i", "--input-freq-mhz", help="PLL Input Frequency", type=float, default=27
+)
+parser.add_argument(
+    "-o", "--output-freq-mhz", help="PLL Output Frequency", type=float, default=108
+)
+parser.add_argument("-d", "--device", help="Device", type=str, default="GW1NR-9 C6/I5")
```
yrabbit, line 19: Most of our tools accept the chip indication parameter as a part number or as a series name. Having two ways is not good in itself, and we would not want to add a third way of specifying it.

Could you extract the information you need from the part number?
GW1N-UV4LQ144C6/I5
GW1N-LV1QN48C6/I5
GW1NZ-LV1QN48C6/I5
GW1NSR-LV4CQN48PC7/I6
GW1NR-LV9QN88PC6/I5

PR #192, opened 2023-08-04
Install gowin_pll tool

### LINKED ISSUE #191: Please export gowin-pll as a script
This enables using it after just installing apicula.

whitequark:
Fixes #191.

I implemented this in a somewhat cursed way where it does `sys.exit(0)` because otherwise the caller (the script installed by setuptools) would crash trying to call the module. This minimizes the diff though.


pepijndevos:
I'd rather wrap it in a main function and add the usual if name == main
like the other scripts

whitequark:
Done

pepijndevos:
Thank you!

PR #196, opened 2023-08-25
Fix IO attributes and Z-1 PLL

yrabbit:
The GW1NZ-1 has PLL power saving enabled by default. We turn it off.

And also we continue to select the formula for setting the OPENDRAIN flag, which, as it turned out, interferes with the processing of PULLMODE


pepijndevos (approved):

yrabbit:
let's wait for confirmation from Catherine?

whitequark:
I can confirm the mismatches with proprietary tools are gone. The netlist I had issues with is translated equally with Apicula and Gowin tools with this patch.

PR #199, opened 2023-09-04
Himbaechel. Improve dedicated clock router

yrabbit:
Added buffers for the global clock network.
We indicate that the network is CLOCK by placing the following construction in the file of restrictions:

CLOCK_LOC "net_name" BUFG;

As a result, the specified network is cut into two parts: the part from the source to the buffer is routed using any available PIPs, while the part from the buffer to the sink is routed through the global clock network.

This will slightly improve the situation with Tangnano20k, since now it is possible to declare its pin, on which the simple external generator is located, as a clock, which will lead to routing through the buffer without any clock glitches.

All examples for Tangnano20k, which previously used the PLL as a buffer, are no longer needed and use the same files as for the rest of the boards.

!!! Networks connected to dedicated clock pins should not be marked as CLOCK - although this will work, it does not make sense to add a buffer to a network that is already directly connected to the global clock network.


pepijndevos (approved):
Looks good to me!

Except that pesky `Python3_EXECUTABLE` :(

pepijndevos:
How about using https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#python-path

pepijndevos:
I spent too much time playing with Lego today so I didn't get to do any hardware testing. Any particular hardware I have that you'd like to test but haven't?

yrabbit:
> I spent too much time playing with Lego today so I didn't get to do any hardware testing. Any particular hardware I have that you'd like to test but haven't?

as usual attosoc-tec0117 :)

pepijndevos:
Alright, so tec0117, release, milestone

Flinner:
Should the README.md have it listed as supported yet?

yrabbit:
the most recent release of nextpnr from May 16 does not know anything about himbaechel-gowin, so there is no support except in the master branches.

pepijndevos:
The readme does tell people to install the master branch though, so I think it's fine to list it as supported.

PR #201, opened 2023-10-01
Fix build attempt

davidsiaw:
Support for Primer 20k was not added until [June this year](https://github.com/YosysHQ/nextpnr/commit/1260f2f7d7deba283f7c775c6aab8899925b0cf8). This means that the build will fail when used against the nextpnr ver 0.6, which does not have this update.

This causes the entire build to look like it is failing, which is kind of bad and may eventually mask future issues, and is also troublesome to check.

This is an attempt to resolve the issue while still testing it on master and everything else on 0.6, which should pass.

Testing done here https://github.com/davidsiaw/apicula/tree/test-build

### DIFF
--- a/.github/workflows/chipdb.yml
+++ b/.github/workflows/chipdb.yml
@@ -306,7 +306,7 @@ jobs:
         make -j$(nproc)
         sudo make install
         cd ../examples
-        make -j$(nproc) all
+        make -j$(nproc) all NEXTPNR_VERSION=${{ matrix.nextpnr }}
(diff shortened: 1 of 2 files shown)


pepijndevos:
I appreciate the effort but it still fails to build on 0.6. TBH we generally only expect master/master to work since nextpnr releases are just too infrequent to keep working after any significant update on our side. Maybe we should just remove nextpnr release from the build matrix until thing stabilize enough.

PR #202, opened 2023-10-03
Himbaechel. Add initial BSRAM support

yrabbit:
- added support for data generation for BSRAM initialization in a bit image;
- BSRAM Bels are created;
- pROM/pROMX9 primitives with widths of 1, 2, 4, 8, 9, 16, 18 bits are supported.
- start of unpacking - only attributes.


yrabbit:
master nextpnr already has BSRAM support so let's see how the examples compile :)

yrabbit:
Suddenly. The log does not show a single point of error, while the theory is that it is fed up with the constant abc9 crashes, it needs to be checked.
[0_example (master, master).txt](https://github.com/YosysHQ/apicula/files/13470205/0_example.master.master.txt)

pepijndevos:
Amazing milestone!

PR #209, opened 2023-11-15
Add 20k devices to readme

pepijndevos:
Hmmm maybe we need instructions for installing himbachel?


yrabbit:
There are no such instructions even on the nextpnr website :) but we can specify that “in the directory with the sources of the nextpnr master branch” execute a sequence of commands, provided that all dependencies are installed as specified in https://github.com/YosysHQ/nextpnr# prerequisites
```
cmake . -DBUILD_PYTHON=OFF -DARCH="gowin;himbaechel" -DHIMBAECHEL_GOWIN_DEVICES="GW1N-1;GW1NZ-1;GW1N-4;GW1N-9;GW1N-9C;GW1NS-4;GW2A-18;GW2A-18C"
make -j$(nproc)
sudo make install
```

pepijndevos:
Maybe we should add it to nextpnr then, more likely to be correct

Issue #206, opened 2023-11-10
 GW1NR-9  vs  GW1N-9 bit-stream


chili-chips-ba:
The only difference between these two devices is in the package add-ons. FPGA die is the same for both. While **N** package includes only FPGA die, the **NR** is a SIP with SDRAM, PSRAM, or double PSRAM packaging options along with FPGA die.

Since those are essentially "external components", by supporting the **N**, Apicula is also supporting the **NR**. So far, all is nice and dandy. 

But, the problem is that current Apicula doesn't allow specifying **NR** for **--family**. Without it, the **NR** bit-stream is built with **N** ID. When we then try to download Apicula-generated bit-stream using Gowin programmer, it complains about incorrect target. It does not seem it would take much to allow building the **NR** bitstream with the proper NR ID, thus opening the path for mixing and matching Apicula with Gowin proprietary tools.

![NR-vs-N](https://github.com/YosysHQ/apicula/assets/67533663/187cbf12-5882-47b8-a86b-413b03fb4c61)

yrabbit:
What quickly came to my mind without having to duplicate databases with and without R or store multiple ids in the database (which would require adding multiple IDE runs to get different ids at the fuzzing stage) was to add the --device-id parameter to gowin_pack.
This parameter will need to be specified very precisely and accurately, as expected by the vendor bootloader.
The solution is not brilliant and any other patches are welcome, especially in light of the fact that I do not have the opportunity to run vendor things that work directly with hardware.

chili-chips-ba:
The new **--device-id** option would indeed be a lightweight approach for developers, yet a flexible one for users. 

However, we must note that your current classification is misaligned with Gowin's (https://www.gowinsemi.com/en/product/detail/46), hence confusing. It'd be more straightforward if you allowed specifying **--family GW1NR-9C**.

Such family could be virtual, essentially an alias to GW1N-9C in everything by the **--device-id**. 

In that way, if the (ordinary) user does not explicitly specify the **device-id**, you can internally look it up based on **--family** and **--device**. 

Should the (power) user specify **device-id**, you take it as-is, without trying to derive from the other passed options.

<img width="530" alt="gowin-littlebee" src="https://github.com/YosysHQ/apicula/assets/67533663/7504e448-714e-4464-94c8-a929f3f47cca">

jeremyherbert:
I wrote this small script:

```
import json
import os
import subprocess
from collections import defaultdict

def build_family_aliases():
[...]

Issue #208, opened 2023-11-15
Tang nano 20k support


jeremyherbert:
Hi, 

Is anyone working on support for the GW2AR-LV18QN88C8/I7 which is used on the [Tang Nano 20k](https://wiki.sipeed.com/hardware/en/tang/tang-nano-20k/nano-20k.html) ?

I suspect that these two issues may be related:

https://github.com/YosysHQ/apicula/issues/206 (NR is the same as N, but with SDRAM)
https://github.com/YosysHQ/apicula/pull/181 (Same part family, no SDRAM and different package)

Just checking here before I start trying to have a look at it myself.

yrabbit:
Tangnano20k is supported up to the point where it runs this RISC thing called ATTOSOC, look at the Makefile for examples:
https://github.com/YosysHQ/apicula/blob/a4f04e0c481d0ab5feb828a311cca1a76c7b18b6/examples/himbaechel/Makefile.himbaechel#L9-L13

If you want to put things in order with R and NR, then PR is always welcome.

jeremyherbert:
Thanks for the tip, I was looking in the examples folder only, not the himbaechel one. 

I will have a look to see if I can work out something for the R/NR.

pepijndevos:
We should really add it to the readme, we tell people to install from git anyway so I don't think we need to wait for a release

chili-chips-ba:
@juninho99 is stuck on this BRAM-related error. Any tips or advice on how to proceed from here?
<img width="371" alt="bram-error" src="https://github.com/YosysHQ/apicula/assets/67533663/2ef69035-1cfa-4bee-a7ee-b3a776e42569">

yrabbit:
nextpnr says that there are no descriptions for DPX9 primitives in the chip database.
So I'm wondering which version of nextpnr (nextpnr-gowin or himbaechel) you are using and which version.

And I’ll say right away that legacy gowin has no memory support.

chili-chips-ba:
Thank you, that's good input, unfortunately not directly forthcoming from documentation. 
- @juninho99 is now trying it with all the latest versions, and also with himbaechel

chili-chips-ba:
BTW, the design that's giving us this trouble is a game in which CPU writes into memory, while HDMI hardware is reading from it. So, it's dual-port BRAM...
- https://github.com/chili-chips-ba/openXC7-TetriSaraj

yrabbit:
[...]

PR #210, opened 2023-11-16
Add IDCODE override

jeremyherbert:
Some parts have the same "base" family as far as nextpnr is concerned but different IDCODEs (for example, GW1N-9 and GW1NR-9). This PR contains some work to change `gowin_pack.py` to override the IDCODE for the different parts as necessary.

related: https://github.com/YosysHQ/apicula/issues/206

Still to do:

- Add building `family_info.json` to CI
- Allow passing any family to nextpnr, not just the base family
- Actually test it


jeremyherbert:
Assuming that the nextpnr PR gets accepted, this should be OK for review now.

pepijndevos:
What is the intended use for this? We mostly expect people to supply the full part number which we parse into the relevant parts to resolve the correct chip database. I consider passing a shorthand almost a legacy feature or implementation detail, because different parts of the tools care about different levels of detail, so it's better to just have a makefile for your project with the full device ID and pass that to all the different tools.

chili-chips-ba:
... from our perspective, as users, the current need to specify both **--family** and **--device** is indeed a bit confusing and extra work, forcing us to look for definitions of the scope of each. And esp. so if tool _--family_ classification is in the misalignment with Gowin's

jeremyherbert:
For the tang nano 9k, you must pass the family as Gowin apparently shipped two different incompatible silicon dies with the same part number. So it will always be necessary to have the option to pass a family (or a new quasi-part will need to be added which doesn’t exist in the Gowin catalog).

chili-chips-ba:
That's interesting!
You seem to be saying that the `GW1NR-9 QN88P` that comes with TangNano9K may have one of two different FPGA dies inside the package. 
- What exactly are these two unintended variants? 
The more insight you can offer on the topic, the better.

jeremyherbert:
From what I understand, there are two Gowin families that are distinct and incompatible, the GW1N-9 and GW1N-9C. However, you can’t tell what family the part is using the part number, you need to look at the date code on the IC to know. The IDCODEs are also different:

  "GW1N-9": {
    "idcode": "1100581B",
    "base_family": "GW1N-9"
  },
  "GW1N-9C": {
    "idcode": "1100481B",
    "base_family": "GW1N-9C"
  },

It also looks like the same problem may exist for the GW2A-18 and GW2A-18C, though they have the same idcode:

  "GW2A-18": {
    "idcode": "0000081B",
    "base_family": "GW2A-18"
  },
  "GW2A-18C": {
    "idcode": "0000081B",
    "base_family": "GW2A-18C"
  },

Having different base families here means that the hash of the device db in the Gowin tools is different.

chili-chips-ba:
http://cdn.gowinsemi.com.cn/DS117E.pdf
image

Our hunch is that the "C" was their highest speed bin, but they at some point decided not to take that route, and standardize on the somewhat relaxed spec base edition, without "C" suffix. 

**GW1NR-9** 
- PFD range = 3MHz - 400 MHz
- VCO range = 400MHz to **1000MHz**
- CLKOUT range = 3.125MHz to **500MHz** 

**GW1NR-9C**
- PFD range = 3MHz - 400 MHz
- VCO range = 400MHz - **1200MHz**
- CLKOUT range = 3.125MHz to **600MHz** 

Such approach would certainly help them reduce operational expenses and increase yield, which all help keep the customer visible pricing very attractive.

chili-chips-ba:
For completeness, here is the Dual-Port BRAM problem we ran into. This may also have to do with this part number restructuring, such as the feature that existed in GW1NR-9C was marginal in the GW1NR-9 :
  - Design was working fine with DPRAM in Bypass + Read-before-Write mode
  - but crashed (in a way that almost looked like a short) with DPRAM in Pipeline + Read-before-Write

Later on, a newer version of UG285 has eliminated the Read-before-Write mode for DPRAM configuration, along with timing diagrams for it. 
image

chili-chips-ba:
BTW, the **proper part to use for TangNano9K is GW1NR-9C** -- If we build for the GW1NR-9 target, the chip on the board returns IDCODE mismatch. This is all with the latest proprietary tool - Gowin_**V1.9.9_x64_win**. 

https://github.com/YosysHQ/apicula/issues/208#issuecomment-1869962123

PR #228, opened 2024-01-23
switch readme to nextpnr-himbaechel

pepijndevos:


yrabbit:
I was now thinking of adding more corrections to the readme (starting with removing mention of the long-suffering NS-2/honeycomb, references to which have already been cleared from the documentation of Gowin itself :)), but I very quickly realized that this would have to throw out the examples and move everything from the examples/ directory himbaechel is a level higher, and from here it would be nice to clean out the code, leaving only the Himbaechel part, and from here eliminate legacy gowin - in total, this will wait for now :smile: 

honeycomb

pepijndevos:
Yes I think all of those are good ideas. One step at a time.

Issue #224, opened 2024-01-20
Move away from numpy


pepijndevos:
On an Intel Linux machine you can just install numpy binary packages and live happily. However, there has been a slow trickle of requests to get rid of C dependencies that are hard to bundle and compile in various contexts.

This mainly comes from the direction of oss-cad-suite (@mmicko) and yowasp (@whitequark) which are uh somewhat unconventional builds. It's not a problem I personally have, but if it makes it easier to install an FPGA toolchain that seems like a positive thing.

Apicula doesn't actually do any linear algebra, we just use Numpy as a matrix of bits. We use a few packing/unpacking operations and that's it for the most part.

@yrabbit suggested we might try to use https://github.com/wadetb/tinynumpy but I actually think that's not super helpful because we don't need the linear algebra and it doesn't have the bit manipulation stuff.

@whitequark mentioned Glasgow has a bits type that we could steal, which is more in line with what we actually use Numpy for I believe, although only one-dimensional: https://github.com/GlasgowEmbedded/glasgow/blob/main/software/glasgow/support/bits.py

Maybe we can steal code from Glasgow to implement `packbits` in tinynumpy or something like that? Or literally just port everything to plain python lists.

So in conclusion, it would be a welcome change that is fairly mechanical, but would require some work to port/implement our bit manipulation operations.

Our numpy usage:

```
$ git grep "np\."
apycula/bslib.py:    return np.fliplr(np.array(bitmap)), hdr, ftr
apycula/bslib.py:    new_bs = np.vstack((bs, bsram_init))
apycula/bslib.py:    bs = np.fliplr(bs)
apycula/bslib.py:    pad = np.ones((bs.shape[0], padlen), dtype=np.uint8)
apycula/bslib.py:    bs = np.hstack([pad, bs])
apycula/bslib.py:    bs=np.packbits(bs, axis=1)
apycula/bslib.py:        lst, _ = np.histogram(bs, bins=[i for i in range(256)])
apycula/bslib.py:            data=np.packbits(data, axis=1))
apycula/chipdb.py:    template: np.ndarray = None
apycula/chipdb.py:    res = np.zeros((db.height, db.width), dtype=np.uint8)
apycula/fuse_h4x.py:    tile = np.zeros((h, w), np.uint8)#+(255-ttyp)
apycula/fuse_h4x.py:    bitmap = np.zeros((height, width), np.uint8)
apycula/fuse_h4x.py:    res = np.zeros((height, width), dtype=np.uint8)
[...]

PR #230, opened 2024-01-25
Replace numpy arrays with lists.

yrabbit:
Dumb replacement of functions. The functions work only to the extent required, no more.

Creating chip databases (this includes fuzzing etc), packing and unpacking work.

Special moments:
   * you will still need numpy to create the PNG, but the import only happens when that function is called.
   * compression packing needs to be tested.
   * perhaps somewhere in places that I rarely use there are numpy calls.

Addresses https://github.com/YosysHQ/apicula/issues/224

### DIFF
--- a/apycula/bitmatrix.py
+++ b/apycula/bitmatrix.py
@@ -1,0 +1,113 @@
+
+def fliplr(bmp):
+    """
+    Flips the entries in each row in the left/right direction.
+    Returns a new matrix.
+    """
+    return [row[::-1] for row in bmp]
+
+def flipud(bmp):
+    """
+    Reverse the order of elements in each column (up/down).
+    Returns a refence.
+    """
+    return bmp[::-1]
+
+def vstack(bmp_0, bmp_1):
+    """
+    Stack matrices in sequence vertically (row wise).
+    Returns a reference.
+    """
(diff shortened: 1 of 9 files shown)


yrabbit:
It's a horror that flaps in the night, but it works, everything is concentrated in one file and can be improved if desired.

But we probably need to remove numpy from some CI config files or dependencies. I don't know about this.

pepijndevos:
Amazing 🤩 I'll have a closer look at it later

mmicko:
This is great. Thank you for this @yrabbit

pepijndevos (approved):
All that remains is to remove numpy from setup.py

https://github.com/YosysHQ/apicula/blob/8323bef886f2e96ea85fa2c632f0c454dde4c42c/setup.py#L21

We can leave it in the docker file for now, which is harmless.

As long as the basic flow works we'll see if people scream about forgotten places where numpy is used.

mmicko:
@pepijndevos When merged I will update oss-cad-suite scripts, so we can reference people on testing on that as well

yrabbit:
Good old runber strikes back. Let's see what kind of special wires it has...

yrabbit:
Found it. I will need to make changes to nextpnr.

PR #233, opened 2024-01-29
Update workflows

mmicko:
Update to latest version of actions due to Node.js related deprecation in GitHub actions


yrabbit:
Well, all the checks are green, that’s where my knowledge of Node.js ends :)

pepijndevos:
That's also all you need to know I think.

PR #234, opened 2024-02-03
Examples. Describe the Tangnano4k clock pin.

yrabbit:
The point is that the external crystal on this board is soldered to a pin, which is a PLL input, not a clock pin. Therefore, common routing is used.

By specifying that this is a buffered network, we are forcing the router to use global clock wires.


yrabbit:
The fact that the network has become buffered may have a very minor effect in cases where you really need to use this pin as an input for a PLL.

For now we can simply remove CLOCK_LOC "clk" BUFG, but in the future we will need to make a small correction to our global router to catch PLL input pins.

yrabbit:
Also runber, let's wait https://github.com/YosysHQ/nextpnr/pull/1286

pepijndevos (approved):

PR #236, opened 2024-03-17
Add examples with tiny RISCV

yrabbit:
Examples of implementation of lessons 15, 16 and 18 from the Bruno Levy course are added
(https://github.com/BrunoLevy/learn-fpga/blob/master/FemtoRV/TUTORIALS/FROM_BLINKER_TO_RISCV/README.md).

These are parts devoted to the use of inferred BRAM for reading (15), for writing (16) and a complex example with the output of the Malderbrot set via UART (18).  Lessons are implemented for Tangnano4k, Tangnano9k and Tangnano20k boards.

The main function of these three examples is to serve as a canary in the mine, to detect regressions when making changes - like attosoc, but with the possibility of gradually simplifying the design - in particularly difficult cases, you can move along the original lessons from the 15th down to the 1st.

Since no new primitives are used there, there is no point in including them in the unpacking.

It might be worth implementing them for other boards where there are enough LUTs for this riscv.

Also a minor change so that new versions of Python do not swear at regular expressions.

Another purpose of including a small riscv is that it is useful for demonstrating the operation of DSP primitives via UART - when working with 18-bit numbers, the LEDs are no longer enough very quickly.

### DIFF
--- a/apycula/tiled_fuzzer.py
+++ b/apycula/tiled_fuzzer.py
@@ -102,7 +102,7 @@ def tbrl2rc(fse, side, num):
 
 # Read the packer vendor log to identify problem with primitives/attributes
 # returns dictionary {(primitive name, error code) : [full error text]}
-_err_parser = re.compile("(\w+) +\(([\w\d]+)\).*'(inst[^\']+)\'.*")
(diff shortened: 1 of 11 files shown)


pepijndevos (approved):
Ohh very nice

PR #238, opened 2024-03-27
update yosys CI branches

pepijndevos:
It seems like Yosys renamed their main branch, breaking our CI

https://github.com/YosysHQ/yosys/pull/4297


mmicko:
Ah.sorry I missed thia one

PR #241, opened 2024-04-08
Add PLL pads.

yrabbit:
Whenever possible, we arrange the PLL so as to use a direct connection of pins to the inputs of the primitive. The main work happens in nextpnr-himbaechel, here we only prepare data about which pin corresponds to which PLL.

Also added an example for Tangnano4k where two PLLs compete for one input pin. Only one of them can win, but both must function.

### DIFF
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -97,6 +97,9 @@ class Device:
     bottom_io: Tuple[str, str, List[Tuple[str, str]]] = field(default_factory = tuple)
     # simplified IO rows
     simplio_rows: Set[int] = field(default_factory = set)
+    # which PLL does this pad belong to. {IOLOC: (row, col, type, bel_name)}
+    # type = {'CLKIN_T', 'CLKIN_C', 'FB_T', 'FB_C'}
+    pad_pll: Dict[str, Tuple[int, int, str, str]] = field(default_factory = dict)
     # tile types by func. The same ttyp number can correspond to different
     # functional blocks on different chips. For example 86 is the PLL head ttyp
     # for GW2A-18 and the same number is used in GW1N-1 where it has nothing to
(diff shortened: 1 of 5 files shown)


yrabbit:
This is just placement and is compatible in all options, but we are waiting
https://github.com/YosysHQ/nextpnr/pull/1309
after which it is possible that Tec0117 will be able to use its quartz at 100 MHz :)

pepijndevos:
If you've tested this on the boards that you have access to, feel free to merge. I will get back to testing the DSP and PLL stuff in my project at some point but my availability will get worse before it will get better.

yrabbit:
I checked the operation of PLL pads on a board close to tec0117 - szfpga GW1N-9.
To do this, I corrected the femto-riscv-15 example so that a PLL was used and the MSB LED showed a LOCKED signal.

``` verilog
   wire lock;
   assign led = ~{lock, x10[4:0]};
```

``` verilog
module Clockworks (
    input  wire CLK,
    input  wire RESET,
    output wire clk,
    output wire resetn,
    output lock
);
    parameter SLOW = 0;

    assign resetn = RESET ^ `INV_BTN;
    generate
        if (SLOW != 0) begin
         rPLL #(
            .FCLKIN("50.0"),
            .IDIV_SEL(12), // -> PFD = 3.8461538461538463 MHz (range: 3-400 MHz)
            .FBDIV_SEL(6), // -> CLKOUT = 26.923076923076923 MHz (range: 400-600 MHz)
            .ODIV_SEL(16) // -> VCO = 430.7692307692308 MHz (range: 600-1200 MHz)
        ) pll (.CLKOUTP(), .CLKOUTD(), .CLKOUTD3(), .RESET(1'b0), .RESET_P(1'b0), .CLKFB(1'b0), .FBDSEL(6'b0), .IDSEL(6'b0), .ODSEL(6'b0), .PSDA(4'b0), .DUTYDA(4'b0), .FDLY(4'b0),
            .CLKIN(CLK), // 50.0 MHz
            .CLKOUT(clk), // 26.923076923076923 MHz
            .LOCK(lock)
        );
```

On the board, the external crystal generates 50MHz which is fed to pin 27, but I connected it to the PLL PAD (pin 35):

shot-1


RISCV successfully worked at 27 MHz taken from the PLL output + LOCKED LED is on


https://github.com/YosysHQ/apicula/assets/6075465/7c954b4b-2c80-4961-9d98-28194d267320

( I will store the changed test files in the branch https://github.com/yrabbit/apicula/tree/gw1n9-pll-pads)

PR #257, opened 2024-07-06
Implement power saving primitive BANDGAP

yrabbit:
As the board on the GW1N-1 chip becomes a rarity, its replacement is the Tangnano1k board with the GW1NZ-1 chip. This chip has a unique mechanism for turning off power to important things such as OSC, PLL, etc.

Here we introduce a primitive that allows energy saving to be controlled dynamically.

Nextpnr will add this primitive automatically if we do not use it explicitly - this is done to explicitly connect the control input since its default connection is neither VCC nor GND.

BANDGAP only has an input, but does not have a fuse - it is on all the time. Of course, a reasonable question arises - what are these magic numbers in line 2258-2259 of gowin_pack.py?

I really don’t want to consider these BANDGAP fuses precisely because this primitive is not configurable - and these bits are fuses that appear in the db.shortval[49]['unknown_75'] and
db.logicinfo['unknown_74'] tables , which clearly describe something with decent configuration capabilities. This could be some kind of adjustable voltage stabilizer, which has an input MUX for control signals and one of the combinations connects BANDGAP.

But this is not part of BANDGAP and I think I'll leave it as it is.

A test example has also been added - himbaechel/bandgap.v - when you press the BANDGAP button, it is triggered and turns off the power to the OSC generator and the LED stops blinking.

### DIFF
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -1574,6 +1574,13 @@ def fse_create_gsr(dev, device):
     dev.extra_func.setdefault((row, col), {}).update(
         {'gsr': {'wire': 'C4'}})
 
+def fse_create_bandgap(dev, device):
+    # The cell and wire are found by a test compilation where the BGEN input is
+    # connected to a button - such wires are easily traced in a binary image.
+    if device in {'GW1NZ-1'}:
+        dev.extra_func.setdefault((10, 18), {}).update(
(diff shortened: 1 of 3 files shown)


uis246:
Maybe in few weeks or a month if I will have free time I may try to look into it.

pepijndevos (approved):

PR #258, opened 2024-07-06
HCLK Support

Seyviour:
## Summary

This pull request documents the CLKDIV and CLKDIV2 HCLK bels for the GW1N-9(C) and GW2A-18(C). A follow up pull request will document pinouts for other devices. This PR also provides two examples of HCLK Bel usage (blinky and DVI),  along with documentation on the general structure of HCLK. 

### Examples
1. blinky-clkdiv: A blinky example that shows CLKDIV configured with different DIV_MODES. `make -f Makefile.himbaechel blinky-clkdiv-tangnano9k.fs`
2. DVI: A simple DVI example that 'wipes' a new colour across the screen every few seconds. `make -f Makefile.himbaechel` dvi-example-tangnano9k.fs`

### Named/Renamed Wires
Along with naming new HCLK wires, this PR renames some others. Specifically, the wires previously named `HCLK_OUT` have been renamed to `HCLK_SECTION_IN` to be more descriptive of their apparent function. This renaming does not affect the rest of the current HCLK implementation since the `HCLK_OUT` name still applies to the wires defined to connect out of HCLK.

### Documentation
`doc/hclk.md` provides documentation on the (currently understood) standard structure of HCLK.

### 
- HCLK in other devices (except perhaps the GW1NS-4) are expected to behave like the HCLK in the GW2A-18. A follow up PR will provide the necessary documentation needed to get HCLK working in them.
- A 'sister' PR will also be made to nextpnr for HCLK. 

###
Companion nextpnr PR: https://github.com/YosysHQ/nextpnr/pull/1340


apycula/gowin_pack.py
```diff
@@ -1962,6 +1962,53 @@ def set_osc_attrs(db, typ, params):
...
+    bin_str = re.findall(bin_pattern, str_val)
+    if bin_str:
+        dec_num = int(bin_str[0], 2)
+        return str(dec_num)
+    return None
+    
+
+
+_hclk_default_params ={"GSREN": "false", "DIV_MODE":"2"}
+def set_hclk_attrs(db, params, num, typ, cell_name, device):
```
yrabbit, line 1976: there is no need to pass device here - this is already a global variable

pepijndevos:
At first glance this looks great, but after the move all my electronics are still in boxes so it's hard for me to test.
So if @yrabbit has reviewed and tested it, the nextpnr side is in, and CI passes, it's good to go as far as I'm concerned.

apycula/gowin_pack.py
Seyviour: That's true! I'll correct that.

Seyviour:
Thank you @pepijndevos. There's a bug on the nextpnr side that I need to address first.

pepijndevos:
Thank you very much!

Seyviour:
My pleasure. Thanks for merging!

PR #259, opened 2024-07-13
Fix regex

max-kudinov:
Fixed regex for shorting device name, so it would work with GW2A series.

Also renamed `luts` to `num`, because this number doesn't necessarily represent number of luts (18 for 20k LUTs chip)

### DIFF
--- a/apycula/gowin_pack.py
+++ b/apycula/gowin_pack.py
@@ -2740,11 +2740,12 @@ def main():
         is_himbaechel = True
 
     # For tool integration it is allowed to pass a full part number
-    m = re.match("GW1N(S|Z)?[A-Z]*-(LV|UV|UX)([0-9])C?([A-Z]{2}[0-9]+P?)(C[0-9]/I[0-9])", device)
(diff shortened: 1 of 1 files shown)


pepijndevos:
Great work!

Maybe we should update the makefile with the chip name to test this change?

https://github.com/YosysHQ/apicula/blob/f19de5230dc1f88c36b6d60f38ae874f1e697ad4/examples/himbaechel/Makefile.himbaechel#L215

max-kudinov:
> Great work!
> 
> Maybe we should update the makefile with the chip name to test this change?
> 
> https://github.com/YosysHQ/apicula/blob/f19de5230dc1f88c36b6d60f38ae874f1e697ad4/examples/himbaechel/Makefile.himbaechel#L215

It seems that this makefile uses short names for all chips, including GW1N*. In that case I think that all names should be changed to their full version, not just GW2A. What do you think about it?

And by the way, what's the difference between `examples` and `examples/himbaechel` directories?

pepijndevos:
Eh that's fair. It could be a useful change, but could be a separate PR.

The difference is example uses nextpnr-gowin while examples/himbaechel uses that.

max-kudinov:
Then the names should be changed in both `examples` and `examples/himbaechel` makefiles, cause `gowin_pack` call is the same. I can make a separate PR with it.

PR #261, opened 2024-07-20
Add DHCEN primitive.

yrabbit:
Added parts necessary to generate chip databases with DHCEN support for all supported boards. The packing was made with the necessary fuses set.

DHCEN is provided as a wire to disable the input MUXes of HCLK, effectively turning off everything that is "beyond" those MUXes.

Most of the work happens in the nextpnr part.


yrabbit:
For now as a draft for two reasons:

 - Tangnano9k has his own ideas about how DHCEN works.
 - There are no examples. Simple things like ODDR & Co have a dual clock input - HCLK and regular clock wires and coming up with a mechanism to ensure HCLK is engaged so that you can use DHCEN doesn't make much sense when there is CLKDIV, so wait for CLKDIV, after which there will be examples.

yrabbit:
In the example, DHCEN controls all networks at once depending on the state of the button. In the unpressed position, all HCLK networks are disabled (in fact, all used input MUXes are disabled) and only the first LED flashes. When pressed, the second LED begins to show the operation of CLKDIV/CLKDIV2.


https://github.com/user-attachments/assets/366d21a4-ba54-483b-ad98-a5721b3e4a1b


https://github.com/user-attachments/assets/49d0990a-2313-4cba-b088-7011d04a6217

pepijndevos:
Seems like this is waiting for https://github.com/YosysHQ/nextpnr/pull/1349 ?

PR #263, opened 2024-07-24
HCLK. Primary clock pins

yrabbit:
We add the ability to use a direct connection of the clock pins on the chip side with the HCLKs input on the corresponding side for the GW1N-9C.

For other chips this standard feature is automatically supported, but 9C is unique - here the wires are renamed.

This is not supported on the bottom side of the chip yet and probably won't be in the near future - but this is not a problem since the external clock generator on the Tangnano9k board is connected to the right side of the chip.


pepijndevos:
How does this relate to #258 ?

yrabbit:
This adds another signal source for the HCLK input MUX and therefore for the primitives that lie behind it, including the CLKDIV2/CLKDIV primitives introduced in the above PR 258.

PR #266, opened 2024-07-26
Add wire class info

yrabbit:
Allow using the architecture for calculating Fmax. Further work will consist of refining the wire classes and the values of time characteristics.


pepijndevos (approved):
Is this for adding back timing driven synthesis in nextpnr-himbaechel?

PR #275, opened 2024-08-29
Remove the magic numbers

yrabbit:
Bring to a unified way of finding out the functions of cells.


pepijndevos (approved):
I'm all for removing magic numbers!

PR #279, opened 2024-09-14
Implement EMCU primitive

yrabbit:
In this PR we add the ARM Cortex M3 microcontroller integrated in the GW1NSR-4C chip that powers the Sipeed TangNano 4K.

This primitive represents the microprocessor itself, but you need to add ROM (flash), RAM, and some I/O to it for proper operation.

Luckily we have working flash, BSRAM and IOBUFs :)

The examples will show how to use this EMCU in the simplest cases.


yrabbit:
I'll add a couple more examples, but actually they won't go through compilation yet - wait for 
https://github.com/YosysHQ/yosys/pull/4594

yrabbit:
It blinks with the LED that on the board, MODE but what can you do.

https://github.com/user-attachments/assets/3179670f-e9c6-4187-bf2c-29ef21f5d2b0

yrabbit:
adv

yrabbit:
We not only flash the LED, but also read its status - it is a read/write bus after all. :)


https://github.com/user-attachments/assets/f8328deb-3359-4998-a90d-d0a6b6e75f45

read-apb

yrabbit:
Here we demonstrate the implementation of a more advanced peripheral to the Tangnano4k embedded processor - the SPI controller as a device on the APB.
It is very simple but nevertheless it is represented in the address space of the processor as status, control and data registers.

https://github.com/user-attachments/assets/e1938930-d593-4da2-89ed-c3c32cccc646

PR #280, opened 2024-09-20
add magic sdram pin constraints

pepijndevos:
While we have documented the sdram pins of some devices we lack the ability to automatically place them like the vendor tools do. For better compatibility and user friendliness this PR adds the ability to fuzz these magic pin locations.


pepijndevos:
Todo: add all the LiteX supported boards with known pins, and actually test that they work.

PR #286, opened 2024-10-22
Fix SyntaxWarning

eamanu:
This commit remove an invalid escape sequence.

Error originally reported here [0] 

[0] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1085325

### DIFF
--- a/apycula/gowin_pack.py
+++ b/apycula/gowin_pack.py
@@ -65,7 +65,7 @@ def sanitize_name(name):
         retname = name[:-4]
     if _verilog_name.fullmatch(retname):
         return retname
-    return f"\{retname} "
+    return fr"\{retname} "
 
 def extra_pll_bels(cell, row, col, num, cellname):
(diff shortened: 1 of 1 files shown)


pepijndevos:
Could you elaborate on the python version compatibility of the old and new syntax?

Sent from Proton Mail Android

-------- Original Message --------

eamanu:
Hi!

Sorry, I don't know what you mean, but  invalid escape sequence were deprecated in py3.6[0]. In python3.12 the DeprecationWarning was replaced by a SyntaxWarning[1].

This commit convert the literal string in a raw.

[0] https://github.com/python/cpython/issues/98401
[1] https://docs.python.org/3/whatsnew/3.12.html#other-language-changes

pepijndevos:
Thanks. I just found my own answer, the fr combination is as old as format strings themselves so there is no backwards compatibility problem. Lgtm!

Sent from Proton Mail Android

-------- Original Message --------

PR #290, opened 2024-11-19
add missing examples to makefile

pepijndevos:
I found out the bandgap and mul18 examples were not being built.

It seems like on GW1N-4 devices the example errors, not sure why.


pepijndevos:
It seems we have an IDDR example for old nextpnr, but not for himbaechel, maybe a matter of copying it over?

yrabbit:
It must have been a glitch.

Let's add IDDR(C).

yrabbit (approved):

PR #293, opened 2024-11-24
Fix PLL limits for GW2A-18 C8/I7

TrAyZeN:
Hello,

This PR sets the PLL limits of GW2A-18 C8/I7 according to [DS226E](https://cdn.gowinsemi.com.cn/DS226E.pdf).


pepijndevos:
thank you!

PR #294, opened 2024-11-29
Iodelay w

yrabbit:
Each IO contains an IODELAY module, providing a total of 128 (0~127)
delays. The delay may be set statically or dynamically.

The delay block is enabled immediately before/after the IO pad.

This commit adds the basic primitive, which works on all boards.

DELAYB primitive will not be implemented because it is found in rare
unsupported chips, and the fate of DELAYA is unknown - it is not
mentioned in documentation.

Delay is formally part of IOLOGIC, but can be used without ddr/oser/ides
- as shown in the iodelay.v example.

The examples blinky-oddr.v and oser4.v are modified to test the new
primitive at the same time.


yrabbit:
IEM got in here, too:( I'll do a clean PR instead.

PR #295, opened 2024-11-30
Add IODELAY.

yrabbit:
Each IO contains an IODELAY module, providing a total of 128 (0~127) delays. The delay may be set statically or dynamically.

The delay block is enabled immediately before/after the IO pad.

Delay is formally part of IOLOGIC, but can be used without ddr/oser/ides as shown in the iodelay.v example.

The examples blinky-oddr.v and oser4.v are modified to test the new primitive at the same time.


pepijndevos:
oh no CI broke :(
but yay IODELAY

yrabbit:
> oh no CI broke :( but yay IODELAY

Yeah, let's wait until Yosys learns to distinguish between chip families and their primitives and then we'll restart CI for all hung PRs to be sure:)

pepijndevos:
Oh I guess they didn't merge that yet huh...

PR #298, opened 2024-12-14
Implement FF in IO blocks.

yrabbit:
There are three registers in IO blocks: IREG, OREG, and TREG. Not in all of them, but only where there is IOLOGIC because these registers are part of it.

Hence the first limitation: if FF is placed in an IO block, the rest of the IOLOGIC cannot be used.

The second limitation is related to the fact that all three registers have common wires CLK, ClockEnable and LocalSetReset and this dictates the types of flipflop's - let's say DFF (in Gowin's implementation) requires a constant connection CE to VCC, so you can not use in the same IO block DFFE, which CE is a signal.

### DIFF
--- a/examples/himbaechel/ioregs.v
+++ b/examples/himbaechel/ioregs.v
@@ -1,0 +1,30 @@
+module top (
+	input clk,
+	input key_i,
+	input rst_i,
+	output [`LEDS_NR-1:0] led
+);
+
+wire key = key_i ^ `INV_BTN;
+wire reset = rst_i ^ `INV_BTN;
+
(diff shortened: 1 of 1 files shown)


uis246:
Isn't OREG a subset of ODDR/OSER modes?

yrabbit:
Maybe at the lowest level Gowin borrows some transistors from ODDR/OSER to make this flipflop, but I have strong doubts about that - there aren't enough IO blocks for this saving to make any difference.

So I think it's a normal separate register whose output is connected to MUX, which by the way doesn't allow to use any other iologic if OREG is used :)

oreg

PR #301, opened 2024-12-21
Restore CUPCDN with a different bit pattern

### LINKED ISSUE #300: Proposal to re-enable ALU mode
If I correctly understand, #78 disables ALU mode 8 (`CUPCDN`), since it has the same `LUT4` bits as mode 1 (`SUB`) and therefore can't be correctly identified by `gowin_unpack`. An alternative option was mentioned, namely defaulting to `SUB`. I would like to propose setting the 4 unused bits (8 through 11) in `CUPCDN` to some value in `gowin_pack` so that `gowin_unpack` can then recover this mode if produced with that tool, and only defaulting to `SUB` if the usual value is found.

I would like your opinion on the matter before sending any changes. ~I would also like your opinion on adding an extra mode implementing the functionality described [at the end of this page](https://github.com/YosysHQ/apicula/blob/master/doc/alu.md), which could complement mode `9` for implementing a multiplier.~

aerkiaga:
This should suffice to get `CUPCDN` working. Closes #300.

### DIFF
--- a/apycula/chipdb.py
+++ b/apycula/chipdb.py
@@ -368,8 +368,8 @@ def fse_luts(fse, ttyp):
             # CDN
             add_alu_mode(mode, bel.modes, lut, "7",     "0101000001011111")
             # CUPCDN
-            # The functionality of this seems to be the same with SUB
-            # add_alu_mode(mode, bel.modes, lut, "8",     "1010000001011010")
+            # We set bits 8 through 11 to make it distinct from SUB
+            add_alu_mode(mode, bel.modes, lut, "8",     "1010111101011010")
             # MULT   INIT="0111 1000 1000 1000"
             #
             add_alu_mode(mode, bel.modes, lut, "9",     "0111100010001000")


pepijndevos:
Thank you!

Ps if you have ideas how to use the alu In new modes I'm all ears. Gowin doesn't even use most of their own modes and yosys only uses addsub. But it's a fun exercise trying to use the hardware for something else.

aerkiaga:
@pepijndevos I was indeed trying some experiments with alus! For the most part, I see using them manually doesn't save a huge deal of space, but `MUL` does (e.g., 33 vs 128 slices for a 5 x 5 multiplier).

pepijndevos:
Another aspect is speed. The carry chain is much faster than "soft" logic. But yes if we can get yosys to synthesis MUL that would be fantastic. (but how to trade off vs DSP...)

aerkiaga:
I'll order a new device soon, would love to lend a hand improving synthesis.

pepijndevos:
I have some spare Tang Nano 1K I think, but you might want something more capable.

PR #303, opened 2025-01-05
Add Module for Tracing Signals

Seyviour:
### Description:

This PR introduces a set of functions for tracing the flow of signals in an FPGA bitstream. It accounts for intertile connections and supports tracing through LUTs and flip-flops. This PR also applies the introduced tracing functions to `find_sdram_pins` allowing us to find pins using only one or two bitstreams.

#### Important Functions:
    get_input_path_dict: Traces and returns connections that emanate from a specified node.
    get_output_path_dict: Traces and returns connections that lead to a specified node.
    get_path_dict: Combines input and output path dictionaries for a given node.
    enumerate_paths: Generates paths between given source and destination nodes.
    get_paths: Accumulates paths between sources and destinations, with optional sorting.

#### Visualization:

The tracing module also includes a visualization function that highlights active tiles in a path, providing a high-level overview of signal flow in the bitstream. This function uses matplotlib, with the import hidden within the function.

#### Example Usage:
```
path_dict = get_path_dict(tile_dict, db, source_node)
paths = get_paths(path_dict, sources, destinations, sort=True)
draw_paths = {i:path for i, path in enumerate(paths)}
tracing.visualize_grid(draw_path, db.rows, db.cols, show=True, save_name="save.jpeg", checkbox=True)
```

#### Codegen :
The `PnR` class in the codegen module was updated to support non-netlist type design files, and default parameters were added to the configuration classes. These changes should not affect any existing scripts.

#### find_sdram_pin Results
The results of using tracing to identify sdram pins are identical to current results, with the exception of the results for the GW1NS-4. I am inclined to think tracing identifies the correct pins (the current results assign the same IO to multiple hyperram ports), but I have not been able to confirm this. The results are in this attached file:  
[compare.txt](https://github.com/user-attachments/files/18310668/compare.txt)

Demo:
A small demo of the visualization function is included in the GIF below.
tracing


pepijndevos:
What I've been doing for tracing signals so far is unpack a bitstream and use the dot command in yosys. Problem with that is unoptimized it's a mess, and optimized it turns to nothing if there is a missing link.

This seems like a useful addition!

pepijndevos:
oh no I didn't get around to testing this and now there is a conflict :(

yrabbit:
> oh no I didn't get around to testing this and now there is a conflict :(

Well, the conflict is small, but it is in the search area of SDRAM pins :wink:

PR #304, opened 2025-01-12
gowin_pack: fix invalid escape sequence

whitequark:
This causes a warning to be printed on stderr when running `apicula.gowin_pack.main()` directly (but not via the CLI).


pepijndevos:
I'm not sure if the intention of this code was to actually return the slash, so maybe it should be \\?

whitequark:
Oh good point!

PR #306, opened 2025-01-27
Implement MIPI and I3C primitives

yrabbit:
Primitives for forming MIPI signals - MIPI_OBUF and MIPI_OBUF_A, as well as primitive for receiving MIPI signals - MIPI_IBUF - are implemented.

These are not emulation, but real MIPI buffers, which are available in chips GW1NR-9C and GW1NSR-4C (respectively Tangnano9k and Tangnano4k boards).

The added examples are not use cases, but rather compilation checks. This is due to the fact that the documentation for the actual protocol of communication with MIPI devices is not open and manufacturers of displays and cameras are very reluctant (actually never) to share the description of registers and commands for their devices.

Also support of buffer primitive for exchange via I3C protocol is added here. This protocol also belongs to the MIPI alliance and is not found in open sources.

However, happy owners of documentation for their MIPI and I3C devices can use these primitives.


yrabbit:
Serializers and deserializers can also be connected to these things, but in the case of Tangnano4k, the lack of CLKDIV will force you to fiddle with general purpose flip flops dividers.

With high-dimension serializers (like OSER8) there is also a problem that IB input passes through one of the inputs of this serializer. Here you have to remember to set MUX depending on which mode (HS or LP) is used :)

PR #307, opened 2025-01-30
out of tree build

pepijndevos:


yrabbit:
this RISC-V worked well both on its own and as a support mechanism for testing DSP examples before.

I'm curious is our way of using BSRAM so crooked or is this an innovation in yosys that requires a different approach to BSRAM? :)

``` txt
2.9. Executing MEMORY_LIBMAP pass (mapping memories to cells).
found attribute 'ram_style = block' on memory top.CPU.RegisterBank, forced mapping to block RAM
mapping memory top.CPU.RegisterBank via $__GOWIN_SDP_
make: *** [Makefile.himbaechel:319: femto-riscv-16-tangnano4k-synth.json] Segmentation fault (сделан дамп памяти)
```

Complete log [c.log](https://github.com/user-attachments/files/18626697/c.log)

pepijndevos:
If it reproduces locally we could bisect the yosys commit that broke it and report upstream.

yrabbit:
done https://github.com/YosysHQ/yosys/issues/4883

PR #308, opened 2025-02-14
Remove examples that use gowin-nextpnr.

yrabbit:
See https://github.com/YosysHQ/nextpnr/pull/1318


yrabbit:
Checking how it will be without legacy nextpnr

PR #314, opened 2025-02-24
Add segmented clock wires.

yrabbit:
Briefly, these wires can be described as connecting directly to the sinks in a rectangle 4 columns wide and at least 10 rows high. The source can be either at the upper end or at the lower end, which is connected by a system of ‘isolation’ of unused ends.

These small nets are useful when the global ones are exhausted or they are not suitable for a given type of sink and/or source - these wires are omnivorous.


yrabbit:
The fact that there are no errors during master-master compilation does not allow us to relax - it's just that I try to maintain compatibility of both new bases with old nextpnr and old bases with new nextpnr :smile: 

So we wait for https://github.com/YosysHQ/nextpnr/pull/1456  :wink:

PR #320, opened 2025-03-25
Remove legacy chip databases

yrabbit:


yrabbit:
Once you remove the legacy wires, the examples will compile and work fine, but unpacking may stall. We will fix this further.

PR #324, opened 2025-04-04
Fixed bitstream compression, tested on tangnano 9k hardware

regymm:
Bitstream compression (gowin_pack -c) seems broken after numpy is removed. This fixes it. 

The histogram version was broken because bs is actually a 2D array. Passing [y for x in bs for x in y] as a parameter would also work, but I just used built-in itertools for a (marginally) slight performance gain.

### DIFF
--- a/apycula/bitmatrix.py
+++ b/apycula/bitmatrix.py
@@ -90,7 +90,8 @@ def histogram(lst, bins):
     """
     l_bins = len(bins) - 1
     r_lst = [0] * l_bins
-    for val in lst:
+    from itertools import chain
+    for val in chain.from_iterable(lst):
         for i in range(l_bins):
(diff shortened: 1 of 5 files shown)


yrabbit:
Good catch.

Try replacing all 'gowin_pack' with 'gowin pack -c' in examples/Makefile to check. Many files will be made there, including ones for Tangnano9k.

pepijndevos:
Yea would be good to run compression on our ci.

yrabbit:
Ok. The main-master variant did not compile. So something doesn't want to cooperate. Have you checked the compilation of the examples locally? Like go to the 'apicula/examples' directory and run 'make clean all' ?

Depending on the nature of the error, we may have to go back in time and figure out what changed in the compression algorithm.

PR #328, opened 2025-05-01
Add BSRAM operation check

yrabbit:
A femto-riscv-memtest example is added that contains riscv and firmware to test memory writes/reads.

Memory is checked after the end of the program and almost up to the maximum allowable address - whole chunks of 64 bytes are checked for simplicity of the program itself.

Implemented for Tangnano4k, Tangnano9k and Tangnano20k with 16K, 48K and 88K BSRAM check respectively.

Firmware source code in examples/riscv-firmware/memtest-src directory.


yrabbit:
The point was to check bytes in total, not visually by pixels of VGA picture or indirectly by operability of Mandelbrot set calculation - there is no guarantee that all bytes work correctly.

In the terminal you can see the write/read result for each byte. Of course, it is not without disadvantages - the area of the program itself and up to 64 last bytes are not checked. Besides, only the primitives generated by the synthesis are checked :)

mem-start-20

mem-end-20

PR #340, opened 2025-05-24
Remove ByteString usages

xtexx:
They have been removed in Python 3.14.

https://github.com/python/cpython/commit/2f4db5a04d6fa7ba5c1c6b82b482dd7ca48f3382


pepijndevos:
Leading edge! Thanks

PR #273, opened 2024-08-25
Initial support for 5A series

AwooOOoo:
- .fse file support
- .dat file support

PR #342, opened 2025-05-29
Switch to new IDE v1.9.10.03

yrabbit:
Based on AwooOOoo's https://github.com/YosysHQ/apicula/pull/273

The move to the new IDE is due to the need to support the tasty new GW5 family chips (Tangprimer25k board for example).

At this stage, however, the task is to provide support for current chips and this is mostly done - all examples compile and work, but there are exceptions:

 - attosoc and femto18 on TangnPrimer20k;
 - femto18 on Tangnano20k.

Very likely related to the intricacies of using BSRAM. We'll see.


yrabbit:
Moved to #345 in order to really compile under new ide

PR #367, opened 2025-08-13
GW5A. Implement routing of clocks.

yrabbit:
Clock wires are added for the GW5A series.

Interesting points:

 - GW5A-25A has four quadrants, just like GW1N-9 or GW2A-18, but with a different column order and spine arrangement.

 - The fuses of the clock MUX are "spread" across the chip cells. This has been observed before, but not on such a scale: one PIP always had fuses in one cell, but in the GW5A series, some of the fuses may be in one cell and some in another. We had to change the way the fuses are set. In principle, it can also work for older series, but research on this has been left for the future.

 - In the new series, there is no way to use the dedicated clock pin as an input to the clock MUX. There are two options for using the clock pin: either "pull" the signal to the input point from the logic to the clock system, or use a high-speed clock system (HCLK).

- When using HCLK as a mechanism for delivering signals from pins to the global clock system in IO blocks, a specific bit is set, and this is normal. What is not normal is that on the underside of the chip, a pin of unknown purpose is also set in a completely different IO block at the same time. For now, we are simulating this behavior.

Added example apycula/examples/gw5a/big-shift.v, which has a huge counter and an equally huge 24-bit shift register. The bits of the shift register are output to PMOD contacts, to which 8-led pmods are supposed to be connected.

Since the order of LEDs may differ from pmod to pmod, it may be necessary to modify big-shift-primer25k.cst.

The counter and register in the example show the operation of DFF in the GW5A series. All types of flip-flops are supported, just as in previous chip series, and in exactly the same quantity—six per cell. However, in the GW5A, there are eight flip-flops in one cell, and the additional flip-flops have limitations that will be investigated in the near future.


yrabbit:
Big shift register in action! :sweat_smile: 


https://github.com/user-attachments/assets/1a8b6efa-ff13-46aa-9441-acc7b87b8e18

PR #369, opened 2025-08-14
Fix python binary distribution.

yrabbit:


pepijndevos:
I think maybe all we need to do is change setuptools==61.0

yrabbit:
Sure! 

It's just PR, not for the merger, but to remind us that this problem exists :)

PR #502, opened 2026-04-20
Improved Handling of SSPI GPIO Overrides for Arora-V. and added .DS_s…

pinnacletechnologysolutionsltd:
# gowin_pack: Allow SSPI/CPU GPIO overrides for GW5A devices

## Description
Currently, gowin_pack raises a hard exception when a user-requested sspi_as_gpio configuration conflicts with the device's chip-database inference. For the GW5A-25A (Arora-V), this is often a false positive, as these pins are frequently required as GPIO for custom board routing (e.g., on the Tang Primer 25K).

This patch:
 1. Converts the hard exception into a descriptive warning.
 2. Respects explicit command-line overrides (args.sspi_as_gpio is not None)
  rather than forcing the database default.
 3. Resolves conflicts where the toolchain implicitly locks pins that the
  user has explicitly requested to use as GPIO.

This fixes builds for the Tang Primer 25K where the clock and configuration pins require manual GPIO assignment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ generic/blinky.png
 generic/blinky.posp
 generic/blinky.vm
 .vscode
+.DS_Store
--- a/apycula/gowin_pack.py
+++ b/apycula/gowin_pack.py
@@ -3531,8 +3531,13 @@ def place(db, tilemap, bels, cst, args, slice_attrvals, extra_slots):
         elif typ == "PINCFG":
             if args.i2c_as_gpio != ('I2C' in parms):
                 raise Exception(f" i2c_as_gpio has conflicting settings in nexpnr and gowin_pack.")
-            if args.sspi_as_gpio != ('SSPI' in parms):
-                raise Exception(f" sspi_as_gpio has conflicting settings in nexpnr and gowin_pack.")
+            if args.sspi_as_gpio is not None:
+                auto_detected = 'SSPI' in parms
+                if args.sspi_as_gpio != auto_detected:
+                    print(f"Warning: SSPI pin configuration mismatch (flag={args.sspi_as_gpio}, detected={auto_detected}). Proceeding with flag override.")
+                else:
+                    # If the user didn't specify, use the inferred detection
+                    args.sspi_as_gpio = ('SSPI' in parms)
         elif typ.startswith("FLASH"):
             pass
         elif typ.startswith("EMCU"):


yrabbit:
Hmm... well, the whole point of the current test is precisely to ensure that the SSPI pin on the TangPrimer25k is used only as a GPIO and nothing else, because a quartz resonator for the clock is soldered onto it, and using it as an SSPI pin isn't a very good idea.

pinnacletechnologysolutionsltd:
Without the override, the Tang Primer 25k cannot boot or clock its internal logic. Could we perhaps gate this check so it remains a hard error by default, but allows an override via a --force flag or a configuration entry?

If the toolchain forces this pin into SSPI mode, the crystal oscillation is suppressed, and the FPGA fabric never receives the clock. It simply will not start.

yrabbit:
There seems to be some kind of communication issue here :)

No matter what value you specify for the SSPI pin, it will always be set to "use as GPIO" for the TangPrimer25k:

https://github.com/YosysHQ/apicula/blob/b2ef92aef61e91cc5313877c537a85ff8d45d549/apycula/gowin_pack.py#L4641-L4644

Another issue is that this must correspond to the command-line switches for the `nextpnr` call, because in this series of chips, the mode in which the pin operates is set not only by fuses during the `gowin_pack` stage, but also by applying voltage to specific inputs—which means that routes must be routed from VCC/VSS to these ports.

And there can be no soft warning if nextpnr is designed to operate in one pin mode, but the user specifies another for `gowin_pack`.

https://github.com/YosysHQ/nextpnr/blob/33e1c9033b6e741fb6f7b46046a751dd973da02b/himbaechel/uarch/gowin/pack.cc#L76-L93

johncurley:
I’ve looked at the gowin_pack.py source code, and I see the logic that automatically sets args.sspi_as_gpio = True for the GW5A-25A. That explains why it's a mandatory setting.

I understand the logic that sets args.sspi_as_gpio and also I understand the SSPI circuit protection within the Gowin family.
 
The issue is that there's a disconnect between this automatic setting and the exception check that immediately follows it. Even when I provide no command-line flags, the toolchain triggers an exception because nextpnr has already processed the design based on the default configuration, leading to a state mismatch that gowin_pack flags as a conflict. 
  
 1. Forcing the setting to True.
 2. Checking if it's True.
 3. Panicking if it detects even a minor inconsistency between the PnR JSON and the automatic forced state.
  
Since this pin (E2) is physically tied to a 50MHz crystal on the Tang Primer 25K, we have no choice but to force it to GPIO mode—if we don't, the board cannot boot. This is based off real world hardware testing not simulation.

To avoid breaking existing support for other boards, I propose we constrain this relaxed error handling specifically to the GW5A-25A device family. Since this family has unique, hard-wired requirements for SSPI pins on boards like
 the Tang Primer 25K, we could implement a conditional check:

```python
if args.sspi_as_gpio != auto_detected:
    if device == 'GW5A-25A':
        print("Warning: SSPI configuration mismatch on GW5A-25A. Proceeding with override.")
    else:
        raise Exception("SSPI configuration mismatch...")
```
This approach keeps the hard safety check intact for all other chips, while providing the necessary escape hatch for the Arora-V series.

yrabbit:
```
Panicking if it detects even a minor inconsistency between the PnR JSON and the automatic forced state.
```

You see, this pin function is configured in two places—first in the nextpnr module, which handles the wiring to the special ports, and then in the gowin_pack module, which handles fuse settings.

And by "special ports," I specifically mean the special PIN_CFG register, where each bit determines whether a particular pin will function as a GPIO or perform another function.

Is everything clear so far?

From this point on, we cannot allow any "minor inconsistency" because the fuses must match what nextpnr has already done.

And as you keep insisting, TangPrimer25k is useless if  not "to force it to GPIO mode—if we don't, the board cannot boot" well, that's exactly why it's set to True, and therefore nextpnr must have the same flag.

johncurley:
I’m withdrawing this PR. 

While I've verified that my local override of the safety check resolves this for the 25K, I understand the desire to keep the upstream toolchain safe for all configurations. I will manage this as a board-specific configuration in my local environment. 

Thank you for the technical context on the pin-fusing logic.

pepijndevos:
Could you maybe share a MWE that demonstrates what is broken on our current design that is fixed by this PR?

pinnacletechnologysolutionsltd:
Hi pepijndevos,

# Minimal MWE

## To demonstrate the issue, here is the MWE:
 
1. Target: Tang Primer 25K (GW5A-LV25MG121NES).
2. Setup: A simple top-level module where sys_clk is mapped to Pin E2 (the
  hard-wired 50MHz oscillator). 
3. The Conflict: When nextpnr-himbaechel routes this clock to the fabric, it
  correctly identifies the pin, but gowin_pack triggers an exception because it
  detects an sspi_as_gpio conflict that the user cannot override. 
4. Reproduction: 
  - Use the following constraint: IO_LOC "sys_clk" E2; 
  - Attempt to build any design that clocks from this pin.
  - gowin_pack throws: Exception: sspi_as_gpio has conflicting settings...
  
Because this pin is physically tied to the oscillator, there is no alternative routing.

## top.v

This module does nothing but drive an LED with a counter.

```verilog
module top (
        input  wire sys_clk,
        output wire led0
    );
        // Simple counter to ensure the clock is used
        reg [25:0] count = 0;
        always @(posedge sys_clk) count <= count + 1;
        
        // Active low LED
        assign led0 = ~count[25];
   endmodule
```

## top.cst

This file contains the clock pin mapping. The gowin_pack exception is triggered because this pin is dual-purpose.

```verilog
// E2 is the system clock on the Tang Primer 25K
IO_LOC  "sys_clk"  E2;
IO_PORT "sys_clk"  IO_TYPE=LVCMOS33 PULL_MODE=NONE;

// LED0 on L6
IO_LOC  "led0"     L6;
IO_PORT "led0"     IO_TYPE=LVCMOS33 DRIVE=8;
```

## Command sequence

```bash
# Synthesis
yosys -p "read_verilog top.v; synth_gowin -family gw5a -top top -json top.json"

# PnR
nextpnr-himbaechel --device GW5A-LV25MG121NES \
--vopt family=GW5A-25A \
--vopt cst=top.cst \
 --json top.json \
 --write top_pnr.json \
 --freq 50
 
# Pack (The conflict happens here)
gowin_pack -d GW5A-25A top_pnr.json -o top.fs
```

The issue is reproducible because the GW5A-25A database in apycula forces sspi_as_gpio = True implicitly, but the gowin_pack logic enforces a strict check that mismatches the router's state. On the Tang Primer 25K, Pin E2 (the oscillator) requires this pin to be configured as GPIO to function at all. I am proposing a patch to convert this hard exception into a warning for the GW5A-25A family."

## Proposed fix

```python3
# Proposed logic:
if args.sspi_as_gpio != auto_detected:
    if device == 'GW5A-25A':
        print("Warning: SSPI configuration mismatch on GW5A-25A. Proceeding with override.")
    else:
        raise Exception("SSPI configuration mismatch...")
```
However I'll leave it in your hands since I'm not entirely sure how to handle this. For me it works locally and I can carry on with my workflow. Thanks for taking the time to have a look.

PR #505, opened 2026-05-13
gowin_pack: Support driving FF XD from SEL

gatecat:
I was looking at placement performance on a big design and one problem that emerged was the constraint that FF D must only be driven from the LUT (so LUT and unrelated FF cannot coexist) was not encoded very well and limiting performance.

However, I don't see a reason for this constraint to exist so I've added support for this and then can remove the limitation altogether. Only tested on the 138k, if this doesn't work on other devices let me know. Would appreciate testing this if this isn't something you have experimented with already.

### DIFF
--- a/apycula/gowin_pack.py
+++ b/apycula/gowin_pack.py
@@ -292,6 +292,7 @@ def get_bels(data):
         yield (cell['type'], int(row), int(col), num,
                 cell['parameters'], cell['attributes'], sanitize_name(cellname), cell)
 
+_ff_sd_0 = set()
 _pip_bels = []
 def get_pips(data):
     pipre = re.compile(r"X(\d+)Y(\d+)/([\w_]+)/([\w_]+)")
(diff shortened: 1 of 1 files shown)


gatecat:
The nextpnr part is in https://github.com/YosysHQ/nextpnr/pull/1725

pepijndevos:
Iirc the lut and dff are hardwired together but you can indeed use one without the other but not both with unrelated bels. I might be misremembering but I think the constraint should be "if the lut output doesn't feed the dff, the dff can't be used for other things" but it's been a while a gw5a is in many ways a completely different beast.

We have hardware in the loop tests now, both gw1n and gw2a. There is a commented out line to make ci use a nextnpr PR so with that you could write a test that ensure this change works on both families.

yrabbit:
Well, DFF can have an independent input in all series—I’ve implemented nextpnr/apicula variants with SEL support.

What stopped me from submitting a PR for this mechanism was the extremely limited options for connecting the SEL input, unlike the LUT inputs—in the image, the B0 PIPs are listed first, followed by SEL0.

After seeing a sharp increase in the failure rate of routing of examples, and looking at the length of the routes in successful cases, I decided it wasn’t worth it—if attosoc suddenly stopped working, I’d shoot myself :)

SEL

gatecat:
Alright, fair enough, this has never been a problem for ECP5 which has a similar configuration but I'll leave this for now and look at encoding the problem in a different way.

yrabbit:
I didn't mean to say we should abandon it entirely—if you do this only via a command-line switch in nextpnr, rather than by default, then I have no objections at all :)

PR #525, opened 2026-08-07
gowin_pack: add multiboot address support

Voiris:
Adds a `--multiboot_address` option to `gowin_pack` for setting the SPI flash address of the next bitstream.

I needed this functionality for a multiboot setup and decided to add support for it to `gowin_pack`.

The option updates the existing `0xD2` SPI flash address command in the bitstream header.

Hardware tested on a Tang Nano 20K (GW2A-18C):

- Verified that only the `0xD2` command changes when setting a multiboot address.
- Tested with both compressed and uncompressed bitstreams.
- Verified reconfiguration in both directions between bitstreams stored at `0x000000` and `0x200000` using `RECONFIG_N`.

### DIFF
--- a/apycula/gowin_pack.py
+++ b/apycula/gowin_pack.py
@@ -4592,6 +4592,17 @@ def set_slice_fuses(db, tilemap, slice_attrvals):
             for brow, bcol in bits:
                 tile[brow][bcol] = 1
 
+def set_multiboot_address(header, address):
+    if not 0 <= address <= 0xffffffff:
+        raise ValueError('Multiboot address must fit in 32 bits')
+
+    for command in header:
+        if command[0] == 0xd2:
+            command[4:8] = address.to_bytes(4, 'big')
+            return
+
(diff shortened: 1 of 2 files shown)


yrabbit:
Very interesting! Is it possible to modify the Makefile in the examples so that two .fs files (blinky and shift, for example) are flashed from different addresses?

Voiris:
Added a Tang Nano 20K multiboot target to the examples Makefile and tested it on hardware.

The example builds two bitstreams:
- `blinky` at `0x000000` with `--multiboot_address 0x200000`
- `shift` at `0x200000` with `--multiboot_address 0x000000`

This allows switching between them using `RECONFIG_N`.

Target: `tangnano20k-multiboot`

yrabbit:
Nice! :+1: 

https://github.com/user-attachments/assets/06b2246e-6d25-49d6-a150-06ab527bb7ce

Voiris:
Glad to see it working 👍

PR #527, opened 2026-08-10
gowin_pack: validate multiboot address alignment

Voiris:
Multiboot addresses must be 4KiB aligned. Only the upper 20 bits of the 32-bit address are used for the multiboot address; the lower 12 bits are reserved and unused.

Currently `gowin_pack` accepts unaligned addresses and successfully generates a bitstream, but the FPGA fails to load the next bitstream during reconfiguration.

This adds a validation check to reject unaligned multiboot addresses.

Hardware tested on a Tang Nano 20K (GW2A-18C):
- An aligned address (`0x200000`) reconfigures successfully.
- An unaligned address (`0x200123`) is accepted by `gowin_pack`, but reconfiguration fails.


yrabbit:
I added this check of yours to the new gowin_pack :)

https://github.com/YosysHQ/apicula/blob/ca88be02ac57a95a75f57ffce030c8e0b670cc4a/apycula/gowin_pack.py#L70-L78

Voiris:
Good rewrite 👍

yrabbit:
I think there will be some bugs, but overall, it's easier to fix (and add new chips): already-established architectures don't break, and there are very few "if then else" situations like "do this, but if it's GW2A, then do that, and if it's GW5A, then..." :)

All examples compile (and support multiboot ;) ).

